PR #37, opened 2024-04-11
add streaming

### LINKED ISSUE #3: Add streaming

aniketmaurya:
<details>

</details>

## What does this PR do?

Fixes #3 

This PR implements support for serving endpoints with streaming support. To enable streaming, you would need to pass `stream=True` in `LitServer` and implement `litapi.predict` and `litapi.encode_response` methods which yields content. 

```py
import time

from typing import Generator

from litserve.api import LitAPI

from litserve import LitServer
from pydantic import BaseModel


class PromptRequest(BaseModel):
    prompt: str


class SimpleStreamAPI(LitAPI):
        self.sentence = "LitServe is streaming output"

    def decode_request(self, request: PromptRequest) -> str:
        return request.prompt

    def predict(self, x) -> Generator:
        output = f"prompt={x} generated_output={self.sentence}".split()
        yield from output

    def encode_response(self, output: Generator) -> Generator:
        for out in output:
            time.sleep(0.1)
            yield out.lower()


if __name__ == "__main__":
    api = SimpleStreamAPI()
    server = LitServer(api, stream=True, timeout=10)
    server.run(port=8888)

```

Example of a client consuming the streamed content:

```py
import requests 

url = "http://127.0.0.1:8888/stream-predict"

resp = requests.post(url, json={"prompt": "how are you"}, headers=None, stream=True)
for line in resp.iter_content():
    if line:
        print(line)
```

- [x] Fix windows tests
- [x] Make `litapi.encode_response` streaming aware
- [x]  concurrent requests
- [x] Stop loop on timeouts
- [x] Added tests

## PR review

## Did you have fun?

### DIFF
--- a/src/litserve/api.py
+++ b/src/litserve/api.py
@@ -57,7 +57,7 @@ class LitAPI(ABC):
 
     @abstractmethod
     def predict(self, x):
-        """Run the model on the input and return or stream the output."""
+        """Run the model on the input and return or yield the output."""
         pass
 
     def unbatch(self, output):
@@ -68,5 +68,9 @@ class LitAPI(ABC):
 
     @abstractmethod
     def encode_response(self, output):
-        """Convert the model output to a response payload."""
+        """Convert the model output to a response payload.
+
+        To enable streaming, it should yield the output.
+
+        """
         pass
--- a/src/litserve/api.py
+++ b/src/litserve/api.py
@@ -45,7 +45,7 @@ class LitAPI(ABC):
 
     @abstractmethod
     def predict(self, x):
-        """Run the model on the input and return the output."""
+        """Run the model on the input and return or stream the output."""
         pass
 
     def unbatch(self, output):
--- a/src/litserve/server.py
+++ b/src/litserve/server.py
@@ -27,7 +27,7 @@ from fastapi import FastAPI, Depends, HTTPException, BackgroundTasks, Request, R
 from fastapi.security import APIKeyHeader
 import sys
 
-from starlette.responses import StreamingResponse
+from fastapi.responses import StreamingResponse
 
 from litserve import LitAPI
 
--- a/src/litserve/server.py
+++ b/src/litserve/server.py
@@ -115,8 +115,9 @@ def run_streaming_loop(lit_api, request_queue: Queue, request_buffer):
             continue
 
         x = lit_api.decode_request(x_enc)
-        for y in lit_api.predict(x):
-            y_enc = lit_api.encode_response(y)
+        y_gen = lit_api.predict(x)
+        y_enc_gen = lit_api.encode_response(y_gen)
+        for y_enc in y_enc_gen:
             with contextlib.suppress(BrokenPipeError):
                 pipe_s.send(y_enc)
 
--- a/src/litserve/server.py
+++ b/src/litserve/server.py
@@ -122,11 +122,9 @@ def run_streaming_loop(lit_api, request_queue: Queue, request_buffer):
         pipe_s.send(StopIteration())
 
 
-def inference_worker(
-    lit_api, device, worker_id, request_queue, request_buffer, max_batch_size, batch_timeout, streaming
-):
+def inference_worker(lit_api, device, worker_id, request_queue, request_buffer, max_batch_size, batch_timeout, stream):
     lit_api.setup(device=device)
-    if streaming:
+    if stream:
         run_streaming_loop(lit_api, request_queue, request_buffer)
         return
     if max_batch_size > 1:
@@ -181,7 +179,7 @@ async def lifespan(app: FastAPI):
                 app.request_buffer,
                 app.max_batch_size,
                 app.batch_timeout,
-                app.streaming,
+                app.stream,
             ),
             daemon=True,
         )
@@ -201,7 +199,7 @@ class LitServer:
         timeout=30,
         max_batch_size=1,
         batch_timeout=0.0,
-        streaming=False,
+        stream=False,
     ):
         if batch_timeout > timeout:
             raise ValueError("batch_timeout must be less than timeout")
@@ -216,7 +214,7 @@ class LitServer:
         self.app.batch_timeout = batch_timeout
         initial_pool_size = 100
         self.max_pool_size = 1000
-        self.app.streaming = streaming
+        self.app.stream = stream
         self.pipe_pool = [Pipe() for _ in range(initial_pool_size)]
 
         decode_request_signature = inspect.signature(lit_api.decode_request)
--- a/src/litserve/server.py
+++ b/src/litserve/server.py
@@ -14,6 +14,9 @@ import uuid
 from fastapi import FastAPI, Depends, HTTPException, BackgroundTasks, Request, Response
 from fastapi.security import APIKeyHeader
 import sys
+
+from starlette.responses import StreamingResponse
+
 from litserve import LitAPI
 
 # if defined, it will require clients to auth with X-API-Key in the header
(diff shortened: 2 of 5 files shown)


aniketmaurya:
blocked by #38

lantiga (commented):
Looks great so far. Adding batching should just work, we just need to loop over the batch at every output (token) generated. We can discuss later about when to best send StopIteration down the pipe for certain items in the batch and not others, but sending when every generation has completed is good enough for now (it’s equivalent to what happens to non streaming batched generation)

src/litserve/server.py
```diff
@@ -99,16 +102,31 @@ def run_single_loop(lit_api, request_queue: Queue, request_buffer):
...
+            try:
+                x_enc, pipe_s = request_buffer.pop(uid)
+            except KeyError:
+                continue
+        except (Empty, ValueError):
+            continue
+
+        x = lit_api.decode_request(x_enc)
+        for y in lit_api.predict(x):
+            y_enc = lit_api.encode_response(y)
```
lantiga, line 118: Shouldn't encode_response be aware of streaming here? We may encode differently at the beginning and end of the stream, I think we should make encode_response a generator like predict in the streaming case.

aniketmaurya: you mean something like @lantiga - 

```py
x = lit_api.decode_request(x_enc)
y = lit_api.predict(x)  # returns a generator
y_enc = lit_api.encode_response(y)  # accepts and returns a generator


##################################
def encode_response(y):
    for item in y:
        yield do_something(item)
```

src/litserve/server.py
```diff
@@ -100,16 +103,32 @@ def run_single_loop(lit_api, request_queue: Queue, request_buffer):
...
-    worker_id,
-    request_queue,
-    request_buffer,
-    max_batch_size,
-    batch_timeout,
-):
+def run_streaming_loop(lit_api, request_queue: Queue, request_buffer):
+    while True:
+        try:
+            uid = request_queue.get(timeout=1.0)
```
aniketmaurya, line 109: enabling batching would be straightforward here as you mentioned in earlier comment @lantiga. Would just have to collect the requests using `collate_request(...)` like [here](https://github.com/Lightning-AI/litserve/blob/a75e342fdc8a3248f72175b161c0c587f8187baf/src/litserve/server.py#L62). 

In this PR, the `streaming_loop` doesn't take any batching argument and hence won't add batching. I would like to enable that in a separate PR if you're okay with that.

src/litserve/server.py
```diff
@@ -295,6 +317,55 @@ async def data_reader():
...
+                return evt.is_set()
+
+            async def stream_from_pipe():
+                while True:
+                    if read.poll(self.app.timeout):
+                        data = read.recv()
+                        if isinstance(data, StopIteration):
+                            return
+                        yield data
+                    asyncio.sleep(0.01)
```
aniketmaurya, line 347: Had to do this for Windows :(. Any suggestion @lantiga?

src/litserve/api.py
```diff
@@ -68,5 +68,9 @@ def unbatch(self, output):
 
     @abstractmethod
     def encode_response(self, output):
-        """Convert the model output to a response payload."""
+        """Convert the model output to a response payload.
+
+        To enable streaming, it should yield the output.
```
lantiga, line 73: in `LitServer` we should check that these functions are actually both yielding using
`inspect.isgeneratorfunction`
when `streaming=True`, and they are not we should return a nice error message

src/litserve/server.py
lantiga: Ok let's keep it for later, but we need to produce an error message if one enables both batching and streaming. Also please add a corresponding test.

lantiga: Also please add open an issue for batching+streaming as part of addressing the review

src/litserve/server.py
lantiga: It is what it is. But let's make sure it's clear that this codepath will only be followed on windows (add a comment or assert outside the while loop)

lantiga: In any case 10ms sleep is a lot while streaming, we need to decrease it by an order of magnitude.

tests/test_lit_server.py
```diff
@@ -119,3 +121,14 @@ def test_run():
...
     process.kill()
+
+
+@pytest.mark.asyncio()
+async def test_stream(simple_stream_api):
+    server = LitServer(simple_stream_api, accelerator="cpu", stream=True)
+    expected_output = "prompt=Hello World generated_output=LitServe is streaming output".lower().replace(" ", "")
+
+    async with LifespanManager(server.app) as manager, AsyncClient(app=manager.app, base_url="http://test") as ac:
+        response = await ac.post("/stream-predict", json={"prompt": "Hello World"}, timeout=10)
```
lantiga, line 132: Can we test individual streaming steps, like wouldn't the test pass even if streaming was not happening?

src/litserve/api.py
lantiga: Also please add tests that test this behavior.

tests/test_lit_server.py
```diff
@@ -119,3 +121,14 @@ def test_run():
     assert '{"output":16.0}' in output
     os.remove("client.py")
     process.kill()
+
+
+@pytest.mark.asyncio()
+async def test_stream(simple_stream_api):
```
lantiga, line 127: We need to add tests for when:
- predict fails during streamed generation
- a timeout happens during streaming

lantiga (changes requested):
Looks great, added a few requests and we can merge

src/litserve/server.py
aniketmaurya: Created [here](https://github.com/Lightning-AI/litserve/issues/45)

src/litserve/api.py
aniketmaurya: added the check and test - https://github.com/Lightning-AI/litserve/pull/37/files#diff-274978086d8b4f049cffd0a53c0a2b1634c323f7ae581e87427c3b0c9046716aR174

tests/test_lit_server.py
```diff
@@ -119,3 +122,57 @@ def test_run():
...
+    with pytest.raises(StopIteration, match="exit loop"):
+        run_streaming_loop(fake_stream_api, requests_queue, request_buffer)
+
+    fake_stream_api.predict.assert_called_once_with("Hello")
+    fake_stream_api.encode_response.assert_called_once()
+
+
+def test_litapi_with_stream(simple_litapi, simple_stream_api):
+    with pytest.raises(
+        ValueError, match="Both lit_api.predict and lit_api.encode_response must be generators in streaming mode."
```
lantiga, line 176: Let's be more explicit

```suggestion
        ValueError, match="When `stream=True` both `lit_api.predict` and `lit_api.encode_response` must generate values using `yield`"
```

and add a simple example inline

src/litserve/server.py
```diff
@@ -27,11 +28,16 @@
...
 import sys
+
+from fastapi.responses import StreamingResponse
+
 from litserve import LitAPI
 
 # if defined, it will require clients to auth with X-API-Key in the header
 LIT_SERVER_API_KEY = os.environ.get("LIT_SERVER_API_KEY")
 
+STOP_ITERATION = pickle.dumps(StopIteration())
```
aniketmaurya, line 39: converted the `StopIteration()` object into its pickle format to avoid pickling error from `multiprocessing.Pipe`.

tests/test_lit_server.py
aniketmaurya: Regarding timeout, there are few caveats @lantiga :
* FastAPI seem to send the streaming content to BackgroundTask so we can't raise TimeoutException during streaming. 
* `server = LitServer(simple_stream_api, stream=True, timeout=10)` -> client will be able to read whatever server could produce within those 10 seconds of timeout.

tests/test_lit_server.py
aniketmaurya: added a test to assert all functions are called correctly by unit testing the [streaming_loop](https://github.com/Lightning-AI/litserve/pull/37/files#diff-274978086d8b4f049cffd0a53c0a2b1634c323f7ae581e87427c3b0c9046716aR153)

tests/test_lit_server.py
aniketmaurya: opened the issue [here](https://github.com/Lightning-AI/litserve/issues/46)

PR #49, opened 2024-04-16
document streaming

aniketmaurya:
<details>

</details>

## What does this PR do?

Fixes # (issue).

## PR review

## Did you have fun?

### DIFF
--- a/README.md
+++ b/README.md
@@ -363,7 +363,7 @@ class SimpleStreamAPI(ls.LitAPI):
     def setup(self, device) -> None:
         self.model = lambda x, y: x * y
 
-    def decode_request(self, request: ls.Request) -> float:
+    def decode_request(self, request):
         return request["input"]
 
     def predict(self, x):
--- a/README.md
+++ b/README.md
@@ -332,13 +332,16 @@ Clients are expected to auth with the same API key set in the `X-API-Key` HTTP h
 
 
 <details>
-  <summary>Stream model response</summary>
+  <summary>Stream long responses</summary>
 
 &nbsp;
 
-`LitServer` can stream model outputs for LLMs or any model of your choice.
+LitServer can stream the outputs for LLMs or any model of your choice.
 
-For example, running the API server for a LLM with streaming:
+To enable streaming, you need to implement `LitAPI.predict` and `LitAPI.encode_response` as a generator (a Python
+function which yields output) and set `LitServer(..., stream=True)`.
+
+For example, running the API server for an LLM with streaming:
 
 ```python
 from typing import Generator
--- a/README.md
+++ b/README.md
@@ -336,7 +336,7 @@ Clients are expected to auth with the same API key set in the `X-API-Key` HTTP h
 
 &nbsp;
 
-LitServer can stream the outputs for LLMs or any model of your choice.
+`LitServer` can stream the outputs for LLMs or any model of your choice.
 
 To enable streaming, you need to implement `LitAPI.predict` and `LitAPI.encode_response` as a generator (a Python
 function which yields output) and set `LitServer(..., stream=True)`.
--- a/README.md
+++ b/README.md
@@ -366,7 +366,7 @@ from litserve import LitAPI
 
 class SimpleStreamAPI(LitAPI):
     def setup(self, device) -> None:
-        self.model = lambda x, y: x**y
+        self.model = lambda x, y: x*y
 
     def decode_request(self, request: Request) -> float:
         return request["input"]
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -329,14 +329,67 @@ LIT_SERVER_API_KEY=supersecretkey python main.py
...
+from litserve import LitServer
+from pydantic import BaseModel
+
+
+class PromptRequest(BaseModel):
+    prompt: str
+
+
+class SimpleStreamAPI(LitAPI):
+    def setup(self, device) -> None:
```
aniketmaurya, line 360: any recommendation on model here which is easy to get started with?

lantiga: Good question, we need to keep this minimal.

Instead of an LLM, can we just generate a sequence of random integers (1-10) as the model?

The other question is around `encode_response`: I think you should structure it as individual payloads (look at how OpenAI or ollama does it), we can't just return raw numbers as the payload.

lantiga (changes requested):

README.md
aniketmaurya: updated to return a json payload instead.

README.md
```diff
@@ -339,8 +339,61 @@ LIT_SERVER_API_KEY=supersecretkey python main.py
...
+
+from litserve.api import LitAPI
+import json
+
+from litserve import LitServer
+from pydantic import BaseModel
+
+
+class Request(BaseModel):
+    number: int
```
williamFalcon, line 367: get rid of this... when you document something, don't introduce random unnecessary concepts.

README.md
```diff
@@ -339,8 +339,61 @@ LIT_SERVER_API_KEY=supersecretkey python main.py
...
+
+
+class SimpleStreamAPI(LitAPI):
+    def setup(self, device) -> None:
+        self.num_outputs = 100
+
+    def decode_request(self, request: Request) -> str:
+        return request.number
+
```
williamFalcon, line 376: same thing here... you're introducing and documenting STREAMING not the pydantic API.

README.md
```diff
@@ -339,8 +339,61 @@ LIT_SERVER_API_KEY=supersecretkey python main.py
...
+            yield x**i
+
+    def encode_response(self, output: Generator) -> Generator:
+        for out in output:
+            yield json.dumps({"output": out})
+
+
+if __name__ == "__main__":
+    api = SimpleStreamAPI()
+    server = LitServer(api, stream=True, timeout=30)
```
williamFalcon, line 388: why are you adding the timeout? is it necessary for straming?
again... don't add unnecessary concepts

README.md
```diff
@@ -339,8 +339,61 @@ LIT_SERVER_API_KEY=supersecretkey python main.py
...
+    server.run(port=8888)
```
williamFalcon, line 389: every example we use uses port 8000... why is this different? does streaming require a different port?

again... when writing documentation about X ONLY talk about X... do NOT introduce other concepts or irrelevant details...

williamFalcon (changes requested):

README.md
aniketmaurya: thank you for the feedback. updating the PR accordingly.

README.md
```diff
@@ -341,8 +341,56 @@ LIT_SERVER_API_KEY=supersecretkey python main.py
...
 
 </details>
+
+
+<details>
+  <summary>Stream long responses</summary>
+
+&nbsp;
+
+`LitServer` can stream longer responses, such as LLM generated text.
```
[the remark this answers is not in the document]
lantiga:
```
This is particularly useful when full generations take several seconds, so that users can see outputs as soon as they are produced.
```

README.md
```diff
@@ -341,8 +341,56 @@ LIT_SERVER_API_KEY=supersecretkey python main.py
...
+
+For example, streaming long responses generated over time:
+
+```python
+from typing import Generator
+
+import json
+from litserve import Request
+from litserve import LitServer
+from litserve import LitAPI
```
williamFalcon, line 364: in examples you MUST minimize the number of import lines...

as a brand we want to do:

```
import litserve as ls
```

which removes 3 import lines because now you can do:
```
ls.Request
ls.LitServer
ls.LitAPI
```

and the top imports can be made a single line:

```
import typing.Generator as Generator, json
```

the result is:
```
import typing.Generator as Generator, json
import litserve as ls
```

williamFalcon (changes requested):
reduce the imports by following my guide

README.md
aniketmaurya: thank you! applied the feedback. I also removed `Generator` type hint since that's not really required to understand what is happening during streaming.

lantiga (approved):
LGTM

PR #50, opened 2024-04-18
add exception handling for LitAPI methods

### LINKED ISSUE #22: handle exceptions from `LitAPI` in the `run_inference` process
## 🐛 Bug

Need to handle user logic for exception. Right now, if the LitAPI methods raise any exception then server doesn't handle it and fails silently. The side effect is that no new request will be processed. 

```py
# Pseudo code

def inference_loop():
    while True:
         uid = get_uid()
        ######## User logic #########
         x = lit_api.decode_request(x_enc)
        y = lit_api.predict(x)
        y_enc = lit_api.encode_response(y)
        ###########################
        with contextlib.suppress(BrokenPipeError):
            pipe_s.send(y_enc)

```

### To Reproduce

1. Run '....'
1. Scroll down to '....'
1. See error

### Environment

- PyTorch Version (e.g., 1.0):
- OS (e.g., Linux):
- Python version:
- CUDA/cuDNN version:

aniketmaurya:
<details>

</details>

## What does this PR do?

Fixes #22 

## PR review

## Did you have fun?

### DIFF
--- a/src/litserve/server.py
+++ b/src/litserve/server.py
@@ -39,6 +39,19 @@ LIT_SERVER_API_KEY = os.environ.get("LIT_SERVER_API_KEY")
 STOP_ITERATION_PKL = pickle.dumps(StopIteration())
 
 
+class LitAPIStatus:
+    OK = "OK"
+    ERROR = "ERROR"
+
+
+def load_and_raise(response):
+    try:
+        exception = pickle.loads(response)
+        raise exception
+    except pickle.PickleError:
+        logging.error(f"Expected response to be a pickled exception but was: {response}")
+
+
 def get_batch_from_uid(uids, lit_api, request_buffer):
     batches = []
     for uid in uids:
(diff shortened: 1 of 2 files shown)


src/litserve/server.py
```diff
@@ -319,7 +342,7 @@ async def event_wait(evt, timeout):
             def get_from_pipe():
                 if read.poll(self.app.timeout):
                     return read.recv()
-                return HTTPException(status_code=504, detail="Request timed out")
+                raise HTTPException(status_code=504, detail="Request timed out")
```
aniketmaurya, line 345: we can just raise the `HTTPException` here and FastAPI will handle things for us.

src/litserve/server.py
```diff
@@ -332,18 +355,18 @@ async def data_reader():
...
 
             if sys.version_info[0] == 3 and sys.version_info[1] >= 8 and sys.platform.startswith("win"):
                 data = await asyncio.to_thread(get_from_pipe)
             else:
                 data = await data_reader()
-
             self.dispose_pipe(read, write)
-            if type(data) == HTTPException:
```
aniketmaurya, line 343: no need for this since we raised the exception

lantiga (changes requested):
Looks good. Added a nit around the error message.

We are missing a test when the exception is generated while streaming right?

aniketmaurya:
@lantiga we have a [test](https://github.com/Lightning-AI/litserve/blob/9f0c054d1c1eeb90fd882fb6988404ad0f4f207e/tests/test_lit_server.py#L153) for streaming loop which asserts to check the yielded output and asserts the predict and encode response but it's not end-to-end test.

lantiga (approved):
Looks good, let's merge this one

Issue #45, opened 2024-04-15
Enable batching with streaming


aniketmaurya:
Also please add open an issue for batching+streaming as part of addressing the review

_Originally posted by @lantiga in https://github.com/Lightning-AI/litserve/pull/37#discussion_r1566469903_

PR #55, opened 2024-04-22
batched streaming

aniketmaurya:
<details>

</details>

## What does this PR do?

Enables batched streaming #45 

- [x] Update tests

## PR review

## Did you have fun?

### DIFF
--- a/src/litserve/server.py
+++ b/src/litserve/server.py
@@ -266,7 +266,7 @@ async def lifespan(app: FastAPI):
 
 
 class LitServer:
-    # TODO: add support for accelerator="auto", devices="auto"
+    # TODO: add support for devices="auto"
     def __init__(
         self,
         lit_api: LitAPI,
--- a/src/litserve/server.py
+++ b/src/litserve/server.py
@@ -106,25 +106,32 @@ def run_single_loop(lit_api, request_queue: Queue, request_buffer):
             pipe_s.send(y_enc)
 
 
-def run_streaming_loop(lit_api, request_queue: Queue, request_buffer):
+def run_streaming_loop(lit_api, request_queue: Queue, request_buffer, max_batch_size, batch_timeout):
     while True:
-        try:
-            uid = request_queue.get(timeout=1.0)
-            try:
-                x_enc, pipe_s = request_buffer.pop(uid)
-            except KeyError:
-                continue
-        except (Empty, ValueError):
+        batches = collate_requests(
+            lit_api,
+            request_queue,
+            request_buffer,
+            max_batch_size,
+            batch_timeout,
+        )
+        if not batches:
             continue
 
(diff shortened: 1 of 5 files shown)


src/litserve/server.py
```diff
@@ -269,6 +305,38 @@ def encode_response(self, outputs):
              """
             )
 
+        if (
+            stream
+            and max_batch_size > 1
+            and not all([
+                inspect.isgeneratorfunction(lit_api.predict),
+                inspect.isgeneratorfunction(lit_api.encode_response),
+                inspect.isgeneratorfunction(lit_api.unbatch),
```
lantiga, line 314: we need to add a default implementation for `unbatch` in case of streaming

let's set the streaming flag to the API instance through the server, so we can `return` or `yield` accordingly in `unbatch`

src/litserve/server.py
```diff
@@ -153,11 +153,50 @@ def run_streaming_loop(lit_api, request_queue: Queue, request_buffer):
             pipe_s.send((pickle.dumps(e), LitAPIStatus.ERROR))
 
 
+def run_batched_streaming_loop(lit_api, request_queue: Queue, request_buffer, max_batch_size, batch_timeout):
```
lantiga, line 156: let's add a comment with a caveat: just like the batched (non-streaming) loop, this will keep generating until all generations in the batch are done

we'll need "continuous batching" to solve this, let's add an issue about it

lantiga: as a more immediate improvement, we need to find a way to avoid sending a lot of tokens past the last token for a particular item in the batch (i.e. we need to trim past the EOS in `encode_response`, let's open an issue and create an example about it)

lantiga (commented):
Looks great so far

src/litserve/server.py
aniketmaurya: @lantiga if I understand correctly we can either `return` or `yield` otherwise the `inspect.isgenerator` won't be able to identify the correctly.

src/litserve/server.py
aniketmaurya: Yeah, I will check on what we can do on here. I have created an issue for now - https://github.com/Lightning-AI/litserve/issues/57

aniketmaurya:
@lantiga we would also need to document on how are we going to batch requests with different argument like `temperature` etc.

src/litserve/server.py
lantiga: right, we probably need to assign the function to the API as LitServe starts

lantiga (commented):
Hey @aniketmaurya this is looking good! Let's remove the [wip] and get it merged

aniketmaurya:
@lantiga please feel free to approve and merge.

lantiga (approved):

PR #60, opened 2024-04-23
NO CHANGE, DON'T MERGE - test mac ci

aniketmaurya:
image

https://github.com/python/peps/pull/3760


aniketmaurya:
@lantiga seems like macos CI for Python 3.9 is dead.

Borda:
just pin Mac OS version, so not to be latest but 13

PR #82, opened 2024-05-11
Let the user know when "setup" has completed...

### LINKED ISSUE #81: Print "Setup complete" when setup is complete
Right now it's kind of a blackbox when things happen... it's useful to know when the setup method was called... so we know everything is ready to go.

williamFalcon:
Let the user know when "setup" has completed...

closes #81


williamFalcon:
@aniketmaurya @lantiga is this the right place for this?

aniketmaurya:
@williamFalcon yes, but I'd suggest using `print` instead of `log` here. Otherwise, whether it should print or not will depend on the user's default logging level. `print` will make sure that it always prints in the user's terminal.

lantiga (approved):

PR #87, opened 2024-05-14
improve logging

aniketmaurya:
This PR add logs with hints and uses `logging.exception` instead of `logging.error` so that users can get error trace.


src/litserve/server.py
```diff
@@ -123,7 +128,10 @@ def run_batched_loop(lit_api, request_queue: Queue, request_buffer, max_batch_si
                 with contextlib.suppress(BrokenPipeError):
                     pipe_s.send((y_enc, LitAPIStatus.OK))
         except Exception as e:
-            logging.exception(e)
+            logger.exception(
+                "Error occurred while processing batched requests.\n"
+                "Hint: Please ensure your LitAPI implementation is correct."
```
lantiga, line 133: Can we be more informative for the user here?

src/litserve/server.py
```diff
@@ -54,7 +56,9 @@ def load_and_raise(response):
         pickle.loads(response)
         raise HTTPException(500, "Internal Server Error")
     except pickle.PickleError:
-        logging.error(f"Expected response to be a pickled exception, but received an unexpected response: {response}.")
+        logger.exception(
+            f"Expected response to be a pickled exception, but received an unexpected response: {response}."
```
lantiga, line 60: It's quite hard to understand this message if you don't have the full picture. Can we make it easier to parse? Like: Error cannot be communicated from the worker to the main process because it failed to be pickled etc.

src/litserve/server.py
aniketmaurya: updated it to ```logger.exception(
                "LitAPI ran into an error while processing the batched request.\n"              
                "Please check the error trace for more details."
            )```

Failure here means user's litapi implementation had issues and error trace would make that clear.

lantiga (approved):

PR #93, opened 2024-05-17
removes `/stream-predict` endpoint and standardizes to `/predict`

aniketmaurya:
* removes `/stream-predict` endpoint and standardizes to `/predict`


lantiga (approved):
Looks good

PR #94, opened 2024-05-20
Bump version since litserve is different from the last released version

rasbt:
This bumps the version so that we can distinguish it from the last released version on PyPI.


lantiga (approved):
Good catch, merging

PR #96, opened 2024-05-21
Exclude markdown files from pre-commit formatter

rasbt:
Let's see if this fixes the issue with removing the whitespaces in the README etc.


rasbt:
Seems like the formatter didn't kick in, so I think this fixed it @lantiga @aniketmaurya

lantiga (approved):
Yes!

PR #101, opened 2024-05-21
add openai streaming

aniketmaurya:
this PR converts the existing OpenAI Spec to return streaming response

### DIFF
--- a/README.md
+++ b/README.md
@@ -559,7 +559,7 @@ class GPT2LitAPI(ls.LitAPI):
 
     def predict(self, prompt):
         out = self.generator(prompt)
-        return out[0]["generated_text"]
+        yield out[0]["generated_text"]
 
 
 if __name__ == '__main__':
@@ -618,21 +618,17 @@ Here is an example of overriding `encode_response` in `LitAPI`:
 
 ```python
 import litserve as ls
-from litserve.specs.openai import ChatCompletionResponseChoice
+from litserve.specs.openai import ChatMessage
 
 class GPT2LitAPI(ls.LitAPI):
     def setup(self, device):
         self.model = None
 
     def predict(self, x):
-        return {"role": "assistant", "content": "This is a generated output"}
+        yield {"role": "assistant", "content": "This is a generated output"}
 
-    def encode_response(self, output: dict) -> ChatCompletionResponseChoice:
-        return ChatCompletionResponseChoice(
-            index=0,
-            message=ChatMessage(role="assistant", content="This is a custom encoded output"),
-            finish_reason="stop",
-        )
+    def encode_response(self, output: dict) -> ChatMessage:
+        yield ChatMessage(role="assistant", content="This is a custom encoded output")
 
 
 if __name__ == "__main__":
--- a/README.md
+++ b/README.md
@@ -444,7 +444,7 @@ if __name__ == "__main__":
 
 
 <details>
-  <summary>Stream long responses</summaryoutput >
+  <summary>Stream long responses</summary>
 
 &nbsp;
 
--- a/README.md
+++ b/README.md
@@ -610,7 +610,7 @@ You can also customize the behavior of `decode_request` and `encode_response` by
 overriding them in `LitAPI`. In this case:
 
 - `decode_request` takes a `litserve.specs.openai.ChatCompletionRequest` in input
-- `encode_response` returns a `litserve.specs.openai.ChatCompletionResponseChoice`
+- `encode_response` returns a `litserve.specs.openai.ChatMessage`
 
 See the OpenAI [Pydantic models](src/litserve/specs/openai.py) for reference.
 
--- a/README.md
+++ b/README.md
@@ -456,7 +456,6 @@ as a generator (a Python function that yields output).
 For example, streaming long responses generated over time:
 
 ```python
-import json
 import litserve as ls
 
 class SimpleStreamAPI(ls.LitAPI):
@@ -472,13 +471,26 @@ class SimpleStreamAPI(ls.LitAPI):
 
     def encode_response(self, output):
         for out in output:
-            yield json.dumps({"output": out})
+            yield {"output": out}
 
 
 if __name__ == "__main__":
     api = SimpleStreamAPI()
     server = ls.LitServer(api, stream=True)
     server.run(port=8000)
+```
+
+To consume a streaming response, your client should iterate through the response content as follows, otherwise 
+your response content would be concatenated as a byte string.
+
+```python
+import requests
+
+url = "http://127.0.0.1:8000/predict"
+resp = requests.post(url, json={"input": 1.0}, headers=None, stream=True)
+for line in resp.iter_content(5000):
+    if line:
+        print(line.decode("utf-8"))
 ```
 
 &nbsp;
--- a/README.md
+++ b/README.md
@@ -611,7 +611,7 @@ You can also customize the behavior of `decode_request` and `encode_response` by
 overriding them in `LitAPI`. In this case:
 
 - `decode_request` takes a `litserve.specs.openai.ChatCompletionRequest` in input
-- `encode_response` returns a `litserve.specs.openai.ChatMessage`
+- `encode_response` yields a `litserve.specs.openai.ChatMessage`
 
 See the OpenAI [Pydantic models](src/litserve/specs/openai.py) for reference.
 
--- a/README.md
+++ b/README.md
@@ -593,6 +593,7 @@ import requests
 
 response = requests.post("http://127.0.0.1:8000/v1/chat/completions", json={
     "model": "my-gpt2",
+    "stream": False,  # You can stream chunked response by setting this True
     "messages": [
       {
         "role": "system",
--- a/src/litserve/server.py
+++ b/src/litserve/server.py
@@ -139,6 +139,7 @@ def run_streaming_loop(lit_api: LitAPI, lit_spec, request_queue: Queue, request_
     while True:
         try:
             uid = request_queue.get(timeout=1.0)
+            logger.debug("uid=%s", uid)
             try:
                 x_enc, pipe_s = request_buffer.pop(uid)
             except KeyError:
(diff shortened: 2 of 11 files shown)


src/litserve/server.py
```diff
@@ -281,6 +282,8 @@ def __init__(
             raise ValueError("batch_timeout must be less than timeout")
         if max_batch_size <= 0:
             raise ValueError("max_batch_size must be greater than 0")
+        if isinstance(spec, OpenAISpec):
```
lantiga, line 285: we can't special case OpenAISpec here
if we pass a spec, we need to determine if the spec overrides attributes on the LitAPI

src/litserve/specs/openai.py
```diff
@@ -70,22 +70,17 @@ class ChatCompletionResponse(BaseModel):
...
+    logprobs: Optional[dict] = None
     finish_reason: Optional[Literal["stop", "length"]]
 
 
 class ChatCompletionChunk(BaseModel):
-    id: str
-    object: str
-    created: int
+    id: str = Field(default_factory=lambda: f"chatcmpl-{shortuuid()}")
+    object: str = "chat.completion"
```
lantiga, line 82: isn't this `"chat.completion.chunk"`?
see https://platform.openai.com/docs/api-reference/chat/create

src/litserve/specs/openai.py
```diff
@@ -70,22 +70,17 @@ class ChatCompletionResponse(BaseModel):
     usage: UsageInfo
 
 
-class Delta(BaseModel):
-    role: str
-    content: str
-
-
 class StreamingChoice(BaseModel):
```
lantiga, line 73: Let's be consistent with naming: this is `ChatCompletionStreamingChoice`

src/litserve/specs/openai.py
```diff
@@ -130,68 +126,80 @@ def encode_response(self, output: Union[Dict[str, str], List[Dict[str, str]]]) -
...
+        for i, response in enumerate(pipe_responses):
+            choices = []
+            async for choice in response:
+                choice = json.loads(choice)
+                logger.debug(choice)
+                choice = StreamingChoice(**choice)
+                choice.index = i
+                choices.append(choice)
+
+            yield ChatCompletionChunk(model=model, choices=choices, usage=usage, system_fingerprint="").json()
```
lantiga, line 186: Look at the spec: https://platform.openai.com/docs/api-reference/chat/streaming#chat/streaming-usage
By default usage is always `null` unless the message is the last message returned

src/litserve/specs/openai.py
```diff
@@ -130,68 +126,80 @@ def encode_response(self, output: Union[Dict[str, str], List[Dict[str, str]]]) -
...
+
+    async def non_streaming_completion(self, request: ChatCompletionRequest, pipe_responses: List):
+        model = request.model
+        usage = UsageInfo()
+        choices = []
+        for i, response in enumerate(pipe_responses):
+            async for choice in response:
+                choice = json.loads(choice)
+                logger.debug(choice)
+                choice["message"] = choice.pop("delta")
```
lantiga, line 196: you need to arrange the loop differently here, so you concatenate messages from the response generator and assemble a single `ChatCompletionResponseChoice`

src/litserve/specs/openai.py
```diff
@@ -130,68 +126,80 @@ def encode_response(self, output: Union[Dict[str, str], List[Dict[str, str]]]) -
...
-                load_and_raise(response)
+        return await self.non_streaming_completion(request, responses)
 
-            if status != LitAPIStatus.OK:
+    async def streaming_completion(self, request: ChatCompletionRequest, pipe_responses: List):
+        model = request.model
+        usage = UsageInfo()
+        for i, response in enumerate(pipe_responses):
+            choices = []
+            async for choice in response:
```
lantiga, line 179: won't this consume the whole response per choice in the absence of a `yield`?

src/litserve/specs/openai.py
aniketmaurya: yes, concatenating it right now

src/litserve/server.py
lantiga: Ok fair, let's keep things hard-coded in this PR and we'll address it later

src/litserve/specs/openai.py
aniketmaurya: updated to iter with zip. thanks for the hint!

src/litserve/specs/openai.py
```diff
@@ -130,68 +134,88 @@ def encode_response(self, output: Union[Dict[str, str], List[Dict[str, str]]]) -
...
+                chat_msg = ChoiceDelta(**chat_msg)
+                choice = ChatCompletionStreamingChoice(
+                    index=i, delta=chat_msg, system_fingerprint="", usage=usage, finish_reason=None
+                )
+
+                choices.append(choice)
 
-            response = response.model_copy(update={"index": i})
-            choices.append(response)
+            chunk = ChatCompletionChunk(model=model, choices=choices, usage=usage).json()
```
lantiga, line 196: From https://platform.openai.com/docs/api-reference/chat/streaming

```
An optional field that will only be present when you set stream_options: {"include_usage": true} in your request. When present, it contains a null value except for the last chunk which contains the token usage statistics for the entire request.
```

Right now we don't have support for usage, but we should arrange this loop so that the last `yield` is special.
To do that do something like

```python
choices = []
usage = None
async for streaming_response in azip(*pipe_repsonses):
    if choices:
        chunk = ChatCompletionChunk(model=model, choices=choices, usage=None).json()
        yield "data: {chunk}\n\n"
    choices = []
    for i, chat msg ...
        ...
        choices.append(choice)
# this is the last chunk
chunk = ChatCompletionChunk(model=model, choices=choices, usage=usage).json()
yield "data: {chunk}\n\n"
yield "data: [DONE]\n\n"
```

src/litserve/specs/openai.py
```diff
@@ -130,68 +134,88 @@ def encode_response(self, output: Union[Dict[str, str], List[Dict[str, str]]]) -
...
+            logger.debug(chunk)
+            yield f"data: {chunk}\n\n"
+        yield "data: [DONE]\n\n"
 
+    async def non_streaming_completion(self, request: ChatCompletionRequest, pipe_responses: List):
         model = request.model
+        usage = UsageInfo()
+        choices = []
+        for i, streaming_response in enumerate(pipe_responses):
+            msgs = ""
```
lantiga, line 206: this should be a list, just append content here and then obtain a single string with `" ".join(msgs)`

README.md
```diff
@@ -610,29 +610,25 @@ You can also customize the behavior of `decode_request` and `encode_response` by
...
-    def encode_response(self, output: dict) -> ChatCompletionResponseChoice:
-        return ChatCompletionResponseChoice(
-            index=0,
-            message=ChatMessage(role="assistant", content="This is a custom encoded output"),
-            finish_reason="stop",
-        )
+        yield {"role": "assistant", "content": "This is a generated output"}
+
+    def encode_response(self, output: dict) -> ChatMessage:
+        yield ChatMessage(role="assistant", content="This is a custom encoded output")
```
lantiga, line 631: This is nicer, but by returning a `ChatMessage` we lose the ability to specify a `finish_reason`.
Right now it was hard-coded to `stop`, but we'll need to differentiate the cases eventually.

lantiga (commented):
Looks good, a few comments before we merge

src/litserve/specs/openai.py
aniketmaurya: Updated to return a last stop chunk with usage and empty choices.

`choices` field of [last chunk](https://platform.openai.com/docs/api-reference/chat/streaming) seem empty:

```
{"id":"chatcmpl-123","object":"chat.completion.chunk","created":1694268190,"model":"gpt-3.5-turbo-0125", "system_fingerprint": "fp_44709d6fcb", "choices":[{"index":0,"delta":{"role":"assistant","content":""},"logprobs":null,"finish_reason":null}]}

{"id":"chatcmpl-123","object":"chat.completion.chunk","created":1694268190,"model":"gpt-3.5-turbo-0125", "system_fingerprint": "fp_44709d6fcb", "choices":[{"index":0,"delta":{"content":"Hello"},"logprobs":null,"finish_reason":null}]}

....

{"id":"chatcmpl-123","object":"chat.completion.chunk","created":1694268190,"model":"gpt-3.5-turbo-0125", "system_fingerprint": "fp_44709d6fcb", "choices":[{"index":0,"delta":{},"logprobs":null,"finish_reason":"stop"}]}
```

src/litserve/specs/openai.py
```diff
@@ -130,68 +134,93 @@ def encode_response(self, output: Union[Dict[str, str], List[Dict[str, str]]]) -
...
+
+            chunk = ChatCompletionChunk(model=model, choices=choices, usage=usage).json()
+            logger.debug(chunk)
+            yield f"data: {chunk}\n\n"
 
-            response = response.model_copy(update={"index": i})
-            choices.append(response)
+        last_chunk = ChatCompletionChunk(
+            model=model,
+            choices=[ChatCompletionStreamingChoice(index=0, delta=ChoiceDelta(), finish_reason="stop")],
```
lantiga, line 202: right but even in this case we can't have just `index=0`, or anything consuming this will break
we need to create an empty `ChoiceDelta` for every choice (i.e. index=0..len(choices))

lantiga (commented):
Looks good, one comment, and it's missing changes to the README.md (which is now incompatible with what is in here).
Also we need to produce a good error message for when the user `return`s from `predict` instead of `yield`ing ("OpenAISpec requires `predict` to be streaming... etc")

src/litserve/specs/openai.py
aniketmaurya: oh I see your point. so when `request.n>1` we need to have that many empty choices, right? How about - 

```py
choices = [ChatCompletionStreamingChoice(index=i, delta=ChoiceDelta(), finish_reason="stop") for i in range(request.n)]
        last_chunk = ChatCompletionChunk(
            model=model,
            choices=choices,
            usage=usage,
        ).json()
        yield f"data: {last_chunk}\n\n"
        yield "data: [DONE]\n\n"
```

aniketmaurya:
@lantiga added validation to ensure that predict and encode_response are generator along with a guide for users.

lantiga (approved):

lantiga:
Looks good, please fix the test and we merge

PR #103, opened 2024-05-22
[WIP] dynamic stream loop selection

### LINKED ISSUE #99: dynamic inference worker
Allow LitServer to send requests to either streaming or non-streaming inference worker dynamically based on the request.

cc: @lantiga

aniketmaurya:
fixes #99 

* Select streaming or non-streaming loop based on client's `request["stream"]` parameter. 

## Server
```py
# server.py
import logging

import litserve as ls
logging.basicConfig(level=logging.INFO)

# STEP 1: DEFINE YOUR MODEL API
    def setup(self, device):
        # Setup the model so it can be called in `predict`.
        self.model = lambda x: x**2

        # Convert the request payload to your model input.
        input = request["input"]
        stream = request.get("stream", False)
        return input, stream

    def predict(self, x):
        input, stream = x
        def gen():
            yield self.model(input)
        if stream:
            return gen()
        return self.model(input)

        # YEILD or return based on output
        ...


# STEP 2: START THE SERVER
if __name__ == "__main__":
    api = SimpleLitAPI()
    server = ls.LitServer(api, accelerator="auto")
    server.run(port=8000)
```

## Client
```py
import requests

response = requests.post("http://127.0.0.1:8000/predict", json={"input": 4.0, "stream": False})
print(f"Status: {response.status_code}\nResponse:\n {response.text}")

```

EDIT: Updated to remove the `iter_*` API

### DIFF
--- a/src/litserve/api.py
+++ b/src/litserve/api.py
@@ -90,7 +90,11 @@ class LitAPI(ABC):
 
     @abstractmethod
     def predict(self, x):
-        """Run the model on the input and return or yield the output."""
+        """Run the model on the input and return the output."""
+        pass
+
+    def iter_predict(self, x):
+        """Run the model on the input and yield the output."""
         pass
 
     def _unbatch_no_stream(self, output):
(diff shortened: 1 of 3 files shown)


williamFalcon (changes requested):
i need a clear user motivation

PR #108, opened 2024-05-23
fix pipe reuse

### LINKED ISSUE #106: concurrent API requests for OpenAI Spec results in inconsistent results
If client A and B hit the server concurrently, client B gets the response for the request provided by client A.

```
client A -> Request("What is the color of sky?")
client B -> Request("What is the color of apple?")
```

`server.dispose_pipe()` recycles the pipes and client B gets the same `Pipe` object back. 

Also, it seems like [Connections](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.connection.Connection.close) are closed automatically when garbage collected and we can call the `close()` method to close it manually.

cc: @lantiga

aniketmaurya:
fixes #106

### DIFF
--- a/src/litserve/server.py
+++ b/src/litserve/server.py
@@ -291,10 +291,7 @@ class LitServer:
         self.max_batch_size = max_batch_size
         self.timeout = timeout
         self.batch_timeout = batch_timeout
-        initial_pool_size = 100
-        self.max_pool_size = 1000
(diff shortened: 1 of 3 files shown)


lantiga (approved):
Let’s merge it as a P0, but please add an issue to investigate recycling pipes

PR #118, opened 2024-05-28
inject context to share/log values across LitAPI methods

aniketmaurya:
When trying to share request context across decode, predict and encode method, the current way to do this is not clean and breaks batching.  This PR injects a `context` that lives across decode, predict, and encode which can be used to log important details that can be passed on from one method to another. 

**User impact**: `context_kwargs` is an optional argument and users won't be forced to include it while defining a custom LitAPI.

Example, we can track the number of tokens for token usage schema of `OpenAISpec`:

```python
import time
from typing import Optional

import litserve as ls
import logging

class GPT2LitAPI(ls.LitAPI):
    def setup(self, device):
        pass

    def predict(self, x, context_kwargs): 
        context_kwargs["num_tokens"] = 0
        for text in "this is a generated text".split():
            context_kwargs["num_tokens"] += 1
            yield text + " "

    def encode_response(self, output, context_kwargs):
        for e in super().encode_response(output):
           if context_kwargs["num_tokens"] > num_max_tokens:
                context_kwargs["finish_reason"] = "length"
                return
           usage = {"num_tokens": context_kwargs["num_tokens"]}
           # we can encode and return usage here
            yield e


if __name__ == '__main__':
    api = GPT2LitAPI()
    server = ls.LitServer(api, accelerator='auto', spec=ls.OpenAISpec())
    server.run(port=8000)
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -764,7 +764,7 @@ response = requests.post("http://127.0.0.1:8000/v1/chat/completions", json={
 </details>    
 
 <details>
-    <summary>✅ Track context throughout request lifecycle</summary>
+    <summary>✅ Track context throughout the request lifecycle</summary>
 
 &nbsp;
 
--- a/README.md
+++ b/README.md
@@ -763,6 +763,35 @@ response = requests.post("http://127.0.0.1:8000/v1/chat/completions", json={
 ```
 </details>    
 
+<details>
+    <summary>✅ Track context throughout request lifecycle</summary>
+
+&nbsp;
+
+You can pass around and track context throughout a request lifecycle by adding a special argument `context` to the 
+`LitAPI` methods.
+
+```python
+import litserve as ls 
+
+class TrackLitAPI(ls.examples.SimpleLitAPI):
+    def predict(self, x, context):
+        context["input_value"] = x
+        return self.model(x)
+
+    def encode_response(self, output, context):
+        # Retrieve the context from LitPI.predict inside encode_response
+        input_value = context["input_value"]
+        return {"output": input_value * output}
+
+if __name__=="__main__":
+    api = TrackLitAPI() 
+    server = ls.LitServer(api)
+    server.run(port=8000)
+```
+
+</details>
+
 &nbsp;
 
 > [!NOTE]
--- a/src/litserve/server.py
+++ b/src/litserve/server.py
@@ -50,10 +50,10 @@ LIT_SERVER_API_KEY = os.environ.get("LIT_SERVER_API_KEY")
 LONG_TIMEOUT = 100
 
 
-def _inject_meta(meta_kwargs: dict, func, *args, **kwargs):
+def _inject_meta(context_kwargs: dict, func, *args, **kwargs):
     sig = inspect.signature(func)
-    if "meta_kwargs" in sig.parameters:
-        return func(*args, **kwargs, meta_kwargs=meta_kwargs)
+    if "context_kwargs" in sig.parameters:
+        return func(*args, **kwargs, context_kwargs=context_kwargs)
     return func(*args, **kwargs)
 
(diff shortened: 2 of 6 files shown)


williamFalcon:
@aniketmaurya start the PR with a clear user story and why this is needed (now).

"When trying to do XYZ, this is not possible/had issue/whatever. this thing we are adding will enable XYZ and it's important because ABC."

src/litserve/server.py
```diff
@@ -50,6 +50,13 @@
 LONG_TIMEOUT = 100
 
 
+def _inject_meta(context_kwargs: dict, func, *args, **kwargs):
+    sig = inspect.signature(func)
+    if "context_kwargs" in sig.parameters:
```
lantiga, line 55: let's call it `context` instead of `context_kwargs` or `meta_kwargs`
let's also change the signature to `_inject_context` accordingly

lantiga (changes requested):
To clarify, this PR will allow users to avoid antipatterns where context data is passed in the main payload if you need it. With this PR you can keep your payload clean (e.g. a tensor), let it batch automatically etc, and pass context data around.

Looks good overall, injection allows users to not think about context if they don't want to.
However the caveat is with unhappy paths: can you show what happens when a user passes two arguments to `predict`?

```
class SomeAPI(LitAPI):
    def predict(x, y):
        ...
```
this will be common, we need to avoid producing a confusing error from Python. If needed, let's catch this and override the error.

Also:
- let's rename the dict to `context` (see comment)
- we need tests

I'm wondering, can't we just have any kwargs after the first positional argument be context, instead of injecting?

aniketmaurya:
> *  However the caveat is with unhappy paths: can you show what happens when a user passes two arguments to predict?
  ```
  class SomeAPI(LitAPI):
      def predict(x, y):
          ...
  ```
[Right now](https://github.com/Lightning-AI/LitServe/blob/720740f08e7496b5de7100f18d9faf379ecbb441/src/litserve/server.py#L123-L125), we pass the decoded output to predict as it is, we don't differentiate if decode returns list or dict.



> * I'm wondering, can't we just have any kwargs after the first positional argument be context, instead of injecting?

If we unroll the kwargs, the objects like integer would be passed as a value instead of reference and any change in one method would not reflect in the other. 


will add tests soon, wanted to get first round of feedback first :D !

lantiga:
> [Right now](https://github.com/Lightning-AI/LitServe/blob/720740f08e7496b5de7100f18d9faf379ecbb441/src/litserve/server.py#L123-L125), we pass the decoded output to predict as it is, we don't differentiate if decode returns list or dict.

Right, I'm referring to the case where we pass 2 arguments to `predict`, not a list. Try to do that on this PR and look at the stack trace and error message. Is it still clear or has it been affected by the injection magic?

aniketmaurya:
@lantiga added a [test](https://github.com/Lightning-AI/LitServe/pull/118/files#diff-274978086d8b4f049cffd0a53c0a2b1634c323f7ae581e87427c3b0c9046716aR389) to make sure that correct error is shown to users in case of unhappy path - 

```
class SomeAPI(LitAPI):
    def predict(x, y):
        ...
```

README.md
```diff
@@ -763,6 +763,35 @@ response = requests.post("http://127.0.0.1:8000/v1/chat/completions", json={
 ```
 </details>    
 
+<details>
+    <summary>✅ Track context throughout request lifecycle</summary>
```
lantiga, line 767:
```suggestion
    <summary>✅ Track context throughout the request lifecycle</summary>
```

lantiga (approved):
Looks good. We need to consider this an experimental feature, the UX may change in the future based on user feedback.

PR #123, opened 2024-05-31
Track device automatically

### LINKED ISSUE #122: Track device automatically.
Track `self.device = device` automatically as a property of the API class

aniketmaurya:
fixes #122

### DIFF
--- a/src/litserve/server.py
+++ b/src/litserve/server.py
@@ -254,6 +254,7 @@ def inference_worker(
     stream,
 ):
     lit_api.setup(device)
+    lit_api.device = device
     message = f"Setup complete for worker {worker_id}."
     print(message)
     logger.info(message)
--- a/src/litserve/api.py
+++ b/src/litserve/api.py
@@ -14,6 +14,7 @@
 import inspect
 import json
 from abc import ABC, abstractmethod
+from typing import Optional
 
 from pydantic import BaseModel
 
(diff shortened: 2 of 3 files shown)


src/litserve/server.py
```diff
@@ -254,6 +254,7 @@ def inference_worker(
     stream,
 ):
     lit_api.setup(device)
+    lit_api.device = device
```
aniketmaurya, line 257: should we have a single method to do both of these? maybe in future we can have `litapi.trigger_setup` which is mean to be used by LitServer to trigger methods.

lantiga: Sorry I didn't get this, can you clarify?

lantiga (approved):
Looks good. In the future we could remove `device` as an argument of `setup`.
Not that it hurts for now, so let's keep it, but we'll have to consider it as we introduce non-PyTorch examples, since not all frameworks have the concept of device.
And maybe as we get into serving the model from multiple devices using tensor or sequential parallelism.

src/litserve/server.py
aniketmaurya: my bad, I thought that I selected two lines while adding the comment. 
So right now, we call `litapi.setup` and then add set `litapi.device`, which kinda creates two source of truth. 

```
lit_api.setup(device)
lit_api.device = device
```

aniketmaurya: saw your other comment, I think it won't matter when we remove device from setup.

PR #124, opened 2024-06-01
fix: kill process

### LINKED ISSUE #121: during manual local testing, the processes are not killed if the test fails
We need to terminate the processes if test fails for whatsoever reason:

## Current
```python
def test_e2e_default_batching(killall):
    process = subprocess.Popen(
        ["python", "tests/e2e/default_batching.py"],
        stdout=subprocess.DEVNULL,
        stderr=subprocess.DEVNULL,
        stdin=subprocess.DEVNULL,
    )

    time.sleep(5)
    resp = requests.post("http://127.0.0.1:8000/predict", json={"input": 4.0}, headers=None)
    assert resp.status_code == 200, f"Expected response to be 200 but got {resp.status_code}"
    assert resp.json() == {"output": 16.0}, "tests/simple_server.py didn't return expected output"
    killall(process)
```

## Proposed

```py
def test_e2e_default_batching(killall):
    process = subprocess.Popen(
        ["python", "tests/e2e/default_batching.py"],
        stdout=subprocess.DEVNULL,
        stderr=subprocess.DEVNULL,
        stdin=subprocess.DEVNULL,
    )

    time.sleep(5)
    try:
      resp = requests.post("http://127.0.0.1:8000/predict", json={"input": 4.0}, headers=None)
      assert resp.status_code == 200, f"Expected response to be 200 but got {resp.status_code}"
      assert resp.json() == {"output": 16.0}, "tests/simple_server.py didn't return expected output"
  except Exception as e:
    raise e
  finally:  # kill the process before raising the exception
    killall(process)
```

---
_Originally posted by @bhimrazy in https://github.com/Lightning-AI/LitServe/issues/119#issuecomment-2141734005_

bhimrazy:
## What does this PR do?

Fixes #121.

### DIFF
--- a/tests/e2e/test_e2e.py
+++ b/tests/e2e/test_e2e.py
@@ -31,17 +31,19 @@ def e2e_from_file(filename):
                 stdin=subprocess.DEVNULL,
             )
             time.sleep(5)
- 
+
             try:
                 test_fn(*args, **kwargs)
-            except Exception as e:
+            except Exception:
                 raise
             finally:
                 parent = psutil.Process(process.pid)
                 for child in parent.children(recursive=True):
                     child.kill()
                 process.kill()
+
         return wrapper
+
     return decorator
 
 
@@ -215,9 +217,7 @@ def test_openai_parity_with_tools():
 
 @e2e_from_file("tests/simple_server.py")
 def test_run():
-    assert os.path.exists(
-        "client.py"
-    ), f"Expected client file to be created at {os.getcwd()} after starting the server"
+    assert os.path.exists("client.py"), f"Expected client file to be created at {os.getcwd()} after starting the server"
     output = subprocess.run("python client.py", shell=True, capture_output=True, text=True).stdout
     assert '{"output":16.0}' in output, f"tests/simple_server.py didn't return expected output, got {output}"
     os.remove("client.py")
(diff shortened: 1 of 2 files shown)


lantiga (commented):
Great! This was super annoying.

We should consider using a decorator to make it easier to write tests this way: https://stackoverflow.com/questions/43937748/wrap-each-pytest-test-function-into-try-except

lantiga:
Taking this for a spin and merging in a few mins : )

lantiga:
Hey @bhimrazy simplified the solution, can you try out the last commits and see if everything checks out?

bhimrazy:
> Hey @bhimrazy simplified the solution, can you try out the last commits and see if everything checks out?

Sure, @lantiga.

Thanks for the solution. I went through the decorator implementation but got stuck trying to kill the process.

 It seems we have a better and more simplified solution now.
 Thanks again.

bhimrazy:
Hi @lantiga,

One of the test cases was failing because the `openai_request_data` from the fixture was not being passed properly through the decorator-wrapper.

After adding `@wraps` to the wrapper, it seems the issue is fixed.

```
@wraps(test_fn)
def wrapper(*args, **kwargs):
```

I think pytest was probably using some of the metadata from the original function.

lantiga (approved):
Great, ship it!

PR #126, opened 2024-06-04
pre-populate context with OpenAISpec

aniketmaurya:
## User motivation

While using OpenAISpec users have to track the openai ChatCompletion request parameters manually in decode_request and most of the time users will need these parameters. LitServe can automatically pre-populate `context` with the parameters so that user don't have to do it manually in decode_request.

## Before
```python
import litserve as ls

    def setup(self, device):
        self.sentence = ["This", "is", "a", "sample", "response"]
   
    ####### This can be automated ########
    def decode_request(self, request, context):
        context["max_tokens"] = request.max_tokens
        return [el.dict() for el in request.messages]

    def predict(self, prompt, context):
        for count, token in enumerate(self.sentence, start=1):
            yield token
            if count > context["max_tokens"]:
                return

if __name__ == "__main__":
    api = SimpleLitAPI()
    server = ls.LitServer(api, spec=ls.OpenAISpec())
    server.run(port=8000)
```

## After

```python
import litserve as ls

    def setup(self, device):
        self.sentence = ["This", "is", "a", "sample", "response"]

    def predict(self, prompt, context):
        for count, token in enumerate(self.sentence, start=1):
            yield token
            if count > context["max_tokens"]:
                return

if __name__ == "__main__":
    api = SimpleLitAPI()
    server = ls.LitServer(api, spec=ls.OpenAISpec())
    server.run(port=8000)
```

### DIFF
--- a/src/litserve/server.py
+++ b/src/litserve/server.py
@@ -170,6 +170,8 @@ def run_streaming_loop(lit_api: LitAPI, lit_spec, request_queue: Queue, request_
 
         try:
             context = {}
+            if isinstance(lit_spec, OpenAISpec):
+                lit_spec.populate_context(context, x_enc)
             x = _inject_context(
                 context,
                 lit_api.decode_request,
--- a/src/litserve/server.py
+++ b/src/litserve/server.py
@@ -170,7 +170,7 @@ def run_streaming_loop(lit_api: LitAPI, lit_spec, request_queue: Queue, request_
 
         try:
             context = {}
-            if isinstance(lit_spec, OpenAISpec):
+            if hasattr(lit_spec, "populate_context"):
                 lit_spec.populate_context(context, x_enc)
             x = _inject_context(
                 context,
--- a/tests/test_specs.py
+++ b/tests/test_specs.py
@@ -108,7 +108,7 @@ async def test_openai_spec_validation(openai_request_data):
             await manager.shutdown()
 
 
-class PrePoulatedAPI(ls.LitAPI):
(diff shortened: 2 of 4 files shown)


aniketmaurya:
[triage-conflicts](https://github.com/Lightning-AI/LitServe/actions/runs/9366025127/job/25782541844?pr=126#logs) CI seem to be having some issue.

src/litserve/server.py
```diff
@@ -170,6 +170,8 @@ def run_streaming_loop(lit_api: LitAPI, lit_spec, request_queue: Queue, request_
 
         try:
             context = {}
+            if isinstance(lit_spec, OpenAISpec):
```
lantiga, line 173: we need to avoid making this specific to the OpenAISpec. The correct condition here is: if the spec implements populate_context, then we call it.

lantiga (changes requested):
Looks good. Condition needs to be fixed, then README needs to have a reference to this functionality added.

tests/test_specs.py
```diff
@@ -106,3 +106,26 @@ async def test_openai_spec_validation(openai_request_data):
     with pytest.raises(ValueError, match="encode_response is not a generator"):
         async with LifespanManager(server.app) as manager:
             await manager.shutdown()
+
+
+class PrePoulatedAPI(ls.LitAPI):
```
lantiga, line 111:
```suggestion
class PrePopulatedAPI(ls.LitAPI):
```

lantiga (approved):
Good to merge after addressing comments

PR #132, opened 2024-06-09
fix: join chat message content, removing empty space

### LINKED ISSUE #128: OpenAI spec non-streaming response contains multiple `space` tokens
While serving a Llama-3 8B instruct LLM using OpenAI spec, the endpoint returns multiple spaces between words. It is because Llama generates contains space in the generated token. 

We should use empty string to concat [OpenAI non-streaming chat content](https://github.com/Lightning-AI/LitServe/blob/81fffb66c6b8a95562317943e7de77dc1f2a5faf/src/litserve/specs/openai.py#L359-L364)

```
for i, streaming_response in enumerate(pipe_responses):
            msgs = []
            tool_calls = None
            async for chat_msg in streaming_response:
                chat_msg = json.loads(chat_msg)
                logger.debug(chat_msg)
                chat_msg = ChatMessage(**chat_msg)
                msgs.append(chat_msg.content)
                if chat_msg.tool_calls:
                    tool_calls = chat_msg.tool_calls

            # Is " " correct choice to concat with?
            content = " ".join(msgs)
            msg = {"role": "assistant", "content": content, "tool_calls": tool_calls}
```

here is the llama response:
image

bhimrazy:
## What does this PR do?

Fixes #128.

### DIFF
--- a/src/litserve/specs/openai.py
+++ b/src/litserve/specs/openai.py
@@ -356,12 +356,11 @@ class OpenAISpec(LitSpec):
                 chat_msg = json.loads(chat_msg)
                 logger.debug(chat_msg)
                 chat_msg = ChatMessage(**chat_msg)
-                # Is " " correct choice to concat with?
                 msgs.append(chat_msg.content)
                 if chat_msg.tool_calls:
                     tool_calls = chat_msg.tool_calls
 
-            content = " ".join(msgs)
+            content = "".join(msgs)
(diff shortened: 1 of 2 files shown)


aniketmaurya (approved):
thanks for the PR @bhimrazy! LGTM 🚀

lantiga:
Thank you @bhimrazy, merged!

PR #133, opened 2024-06-11
Fix Starlette streaming responses

apaz-cli:
Starlette's `StreamingResponse` doesn't behave as intended when you give it an async generator function. It tries to start a threadpool with a bunch of workers to process the async generator. But of course, tokens are sequential. Each task has to wait on the previous, and the order that they're processed seems to be random, so the tokens come in spurts. Everything is consistent once the threadpool closes, but this is not the intended behavior.

To make Starlette process the tokens in order, wrap the async generator with a synchronous one.

I'm less sure about my implementation though. If this could be done better, let me know.

### DIFF
--- a/src/litserve/server.py
+++ b/src/litserve/server.py
@@ -467,12 +467,13 @@ class LitServer:
     async def data_streamer(self, read: Connection, write: Connection):
         data_available = asyncio.Event()
         while True:
-            if not read.poll(1):
+            # Calling poll blocks the event loop, so keep the timeout low
+            if not read.poll(0.001):
                 asyncio.get_event_loop().add_reader(read.fileno(), data_available.set)
                 await data_available.wait()
                 data_available.clear()
                 asyncio.get_event_loop().remove_reader(read.fileno())
-            if read.poll(1):
+            if read.poll(0.001):
                 response, status = read.recv()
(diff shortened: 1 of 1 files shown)


lantiga:
@apaz-cli can you please post an example of the failure, so we can reproduce?

apaz-cli:
@lantiga 
```py
import time
import litserve as ls

    def setup(self, device):
        pass

        return request

    def predict(self, x):
        for i in range(5):
            time.sleep(1)
            yield str(i)

        for o in output:
           yield {"output": o}

if __name__ == "__main__":
    api = SimpleLitAPI()
    server = ls.LitServer(api, stream=True)
    server.run(port=8000)
```

Test with:
```py
import requests

SERVER_URL = 'http://127.0.0.1:8000/predict'

responses = requests.Session().post(SERVER_URL, json="{}", headers=None, stream=True)

print(responses)
for line in responses.iter_lines():
    print(line)
```

lantiga:
Great, for the records (discussed irl) the actual issue was that polling on the pipe was using a high timeout and async was blocking. The confusing part is that this doesn’t happen consistently, but lowering the timeout (we could probably even omit it from my tests) solves the issue, without resorting to async -> sync -> async workarounds.

lantiga (approved):

PR #140, opened 2024-06-17
remove busy wait from data_streamer

aniketmaurya:
<details>

</details>

## What does this PR do?

`data_streamer` polls the multiprocessing Pipe Connection for new data and busy waiting causes async loop to hang as discovered in #133.

* This PR, removes the busy polling and uses async queue to consume high speed data.

* When speed of LitAPI generator (prediction, let's say a for loop of 1B) then consumption speed of data_streamer fails to keep up that pace and data is lost. This PR makes sure that all the data is consumed in a go whenever it's available.

## PR review

## Did you have fun?

### DIFF
--- a/src/litserve/server.py
+++ b/src/litserve/server.py
@@ -476,24 +476,30 @@ class LitServer:
         data_available = asyncio.Event()
         loop = asyncio.get_event_loop()
         loop.add_reader(read.fileno(), data_available.set)
-        while True:
-            # Calling poll blocks the event loop, so keep the timeout low
-            if not read.poll():
+        try:
+            while True:
                 await data_available.wait()
                 data_available.clear()
-            if read.poll(0.001):
-                response, status = read.recv()
-                if status == LitAPIStatus.FINISH_STREAMING:
+                try:
+                    if read.poll(0.001):
+                        response, status = read.recv()
+                        if status == LitAPIStatus.FINISH_STREAMING:
+                            loop.remove_reader(read.fileno())
+                            return
+                        if status == LitAPIStatus.ERROR:
+                            logger.error(
+                                "Error occurred while streaming outputs from the inference worker. "
+                                "Please check the above traceback."
+                            )
+                            loop.remove_reader(read.fileno())
+                            return
+                        yield response
+                except Exception as e:
+                    logger.error(f"Exception occurred: {e}")
                     loop.remove_reader(read.fileno())
                     return
-                if status == LitAPIStatus.ERROR:
-                    logger.error(
-                        "Error occurred while streaming outputs from the inference worker. "
-                        "Please check the above traceback."
-                    )
-                    loop.remove_reader(read.fileno())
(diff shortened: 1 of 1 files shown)


src/litserve/server.py
```diff
@@ -500,24 +500,40 @@ async def win_data_streamer(self, read, write):
...
-                    return
-                if status == LitAPIStatus.ERROR:
-                    logger.error(
-                        "Error occurred while streaming outputs from the inference worker. "
-                        "Please check the above traceback."
-                    )
-                    return
-                yield response
+
+                while not queue.empty():
```
aniketmaurya, line 522: consumes all the data from `Connection` in one go once `data_available` Event is cleared.

lantiga: very good point

src/litserve/server.py
```diff
@@ -500,24 +500,40 @@ async def win_data_streamer(self, read, write):
...
+                    data_available.set()
+            except Exception as e:
+                logger.error(f"Exception in reader: {e}")
+
+        loop.add_reader(read.fileno(), reader)
+
+        try:
+            while True:
                 await data_available.wait()
                 data_available.clear()
```
lantiga, line 520: couldn't there be a scenario where
- `reader` is looping calls `queue.put_nowait` and `data_available.set()`
- the loop further down unblocks, clears `data_available`
- it consumes all items in the queue, while `reader` is still pushing and setting `data_available` afterwards
- the queue now is empty but `data_available` is set
- the loop loops one last time, `data_available` passes, but there are no more items on the queue

ah ok, in that case `queue` is empty and no harm is done. Just wanted to check :-)

src/litserve/server.py
```diff
@@ -500,24 +500,40 @@ async def win_data_streamer(self, read, write):
...
-        while True:
-            # Calling poll blocks the event loop, so keep the timeout low
-            if not read.poll(0.001):
-                asyncio.get_event_loop().add_reader(read.fileno(), data_available.set)
+        queue = asyncio.Queue()
+        loop = asyncio.get_event_loop()
+
+        def reader():
+            try:
+                while read.poll(0.001):  # Check if there's data available to read
```
lantiga, line 508: I would shorten this time, since with this PR we will just consume all available items on one go later on.

Actually... we shouldn't pass anything to `read.poll` at all here. If we want to sleep we need to use `asyncio.sleep` since it will avoid blocking the event loop.

lantiga (approved):
Looks good, but please address the comments before merging

src/litserve/server.py
aniketmaurya: right, ideally timeout is not required here since we only poll when data is available and when it's not then we break this loop. Removed the timeout!

src/litserve/server.py
aniketmaurya: yes, this is the expected flow.

PR #150, opened 2024-06-21
implement OpenAI token usage

aniketmaurya:
<details>

</details>

## What does this PR do?

RIght now we don't provide a way to track prompt tokens and completion tokens and send a dummy value in the response.

This PR will proposes a way to track and update the usage info for `completion_tokens`, `prompt_tokens` and `total_tokens`. 

### Proposal

Return the `prompt_tokens` and `completion_tokens` along with the generated content and LitServe will automatically update it and send in response.

```python
class OpenAIWithUsage(ls.LitAPI):
    def setup(self, device):
        self.model = None

    def predict(self, x):
        yield {"role": "assistant", "content": "This is a generated output", "prompt_tokens": 5, "completion_tokens": 10}

```

Just added test for the expected flow right now, if you approve @lantiga this proposal then I can update the rest asap.

## PR review

## Did you have fun?

### DIFF
--- a/src/litserve/specs/openai.py
+++ b/src/litserve/specs/openai.py
@@ -341,21 +341,21 @@ class OpenAISpec(LitSpec):
         usage = None
         async for streaming_response in azip(*pipe_responses):
             choices = []
-            usage_info = {}
+            usage_infos = []
             for i, (response, status) in enumerate(streaming_response):
                 if status == LitAPIStatus.ERROR:
                     load_and_raise(response)
                 encoded_response = json.loads(response)
                 logger.debug(encoded_response)
                 chat_msg = ChoiceDelta(**encoded_response)
-                usage_info[i] = UsageInfo(**encoded_response)
+                usage_infos.append(UsageInfo(**encoded_response))
                 choice = ChatCompletionStreamingChoice(
                     index=i, delta=chat_msg, system_fingerprint="", finish_reason=None
                 )
 
                 choices.append(choice)
 
-            chunk = ChatCompletionChunk(model=model, choices=choices, usage=sum(usage_info.values())).json()
+            chunk = ChatCompletionChunk(model=model, choices=choices, usage=sum(usage_infos)).json()
             logger.debug(chunk)
             yield f"data: {chunk}\n\n"
 
--- a/src/litserve/specs/openai.py
+++ b/src/litserve/specs/openai.py
@@ -20,7 +20,7 @@ import time
 import typing
 import uuid
 from enum import Enum
-from typing import AsyncGenerator, Dict, List, Literal, Optional, Union, Generator
+from typing import AsyncGenerator, Dict, List, Literal, Optional, Union, Iterator
 
 from fastapi import BackgroundTasks, HTTPException, Request, Response
 from fastapi.responses import StreamingResponse
@@ -100,6 +100,12 @@ class ChatMessage(BaseModel):
     name: Optional[str] = None
     tool_calls: Optional[List[ToolCall]] = None
     tool_call_id: Optional[str] = None
+
+
+class ChatMessageWithUsage(ChatMessage):
+    prompt_tokens: Optional[int] = 0
+    total_tokens: Optional[int] = 0
+    completion_tokens: Optional[int] = 0
 
 
 class ChoiceDelta(ChatMessage):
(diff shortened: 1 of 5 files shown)


lantiga:
thanks for this!

It's a possible design, and it's simple to understand (simpler than requiring to populate context).

A few considerations:
- since we're talking about a streaming API, should we request the user to provide the total over the generation, or should we return what was generated in that step and we accumulate? I'd rather do the former because we need to make fewer assumptions, users need to do the work themselves but it's more explicit on their side too
- let's make sure everything works out in case of batching (it does, but let's always include batching in the design example)
- let's make returning this optional: e.g. returning only at the end of generation would be fine (this is what OpenAI does too)

aniketmaurya:
> since we're talking about a streaming API, should we request the user to provide the total over the generation, or should we return what was generated in that step and we accumulate? I'd rather do the former because we need to make fewer assumptions, users need to do the work themselves but it's more explicit on their side too


Yes, letting users do it explicitly would be cleaner and less assumptious.


> let's make sure everything works out in case of batching (it does, but let's always include batching in the design example)

Thanks for bringing this, I will add a test for this too. 

> let's make returning this optional: e.g. returning only at the end of generation would be fine (this is what OpenAI does too)

+1

williamFalcon:
great idea @aniketmaurya

src/litserve/specs/openai.py
```diff
@@ -368,5 +403,6 @@ async def non_streaming_completion(self, request: ChatCompletionRequest, generat
             msg = {"role": "assistant", "content": content, "tool_calls": tool_calls}
             choice = ChatCompletionResponseChoice(index=i, message=msg, finish_reason="stop")
             choices.append(choice)
+            usage_infos.append(usage)  # Only use the last item from encode_response
 
-        return ChatCompletionResponse(model=model, choices=choices, usage=usage)
+        return ChatCompletionResponse(model=model, choices=choices, usage=sum(usage_infos))
```
aniketmaurya, line 408: OpenAI sums up usage info for `n` choices.

src/litserve/specs/openai.py
```diff
@@ -258,12 +279,12 @@ def _encode_response(self, output: Union[Dict[str, str], List[Dict[str, str]]])
...
             raise HTTPException(500, error)
-
-        return ChatMessage(**message)
+        usage_info = self.extract_usage_info(message)
+        return {**message, **usage_info}
 
     def encode_response(
         self, output_generator: Union[Dict[str, str], List[Dict[str, str]]], context_kwargs: Optional[dict] = None
-    ) -> ChatMessage:
+    ) -> Generator[Dict, None, None]:
```
lantiga, line 287: This is making the spec less clear, let's create a proper type (doesn't ChatMessage have optional usage info already?)

lantiga (approved):
Looks good

See comment, we need to keep code readable and this PR makes it less so. See if you can fix it.

src/litserve/specs/openai.py
aniketmaurya: usage info is set in the outermost level of ChatCompletionResponse object. If we yield a ChatMessage object then it  is converted into a dict (`{"role":..., "content":...}`) when it flows to the main process through the multiprocessing connection Pipe. 

Would you recommend to do something like - 

```
EncodeGenerator = Generator[Union[Dict, ChatMessage], None, None]

def encode_response(
        self, output_generator: Union[Dict[str, str], List[Dict[str, str]]], context_kwargs: Optional[dict] = None
    ) -> EncodeGenerator:
```

lantiga: This still feels too complex. Can't we add an optional usage section to `ChatMessage`? Or even better
```
Generator[Union[ChatMessage, ChatMessageWithUsage]]
```

where we define `ChatMessageWithUsage` explicitly?

aniketmaurya: yes, we can. one question - when we provide the type hint as `ChatMessage` or `ChatMessageWithUsage` which are Pydantic classes, we also allow users to yield dictionaries. Will this type hint create confusion?

lantiga: This is only for `encode_response` right? So most users won't override it, but at least they will know what flows where by glancing at the code.

aniketmaurya: make sense! thank you for clarifying.

aniketmaurya: updated to `Iterator[Union[ChatMessage, ChatMessageWithUsage]]` since Generator has a verbose type hint `Generator[yield_type, send_type, return_type]`

PR #164, opened 2024-07-05
Multiple Queue architecture for process communication

aniketmaurya:
<details>

</details>

## What does this PR do?

This PR removes the multiprocessing strategy of using `Manager().Dict` and Pipes for communication. It introduces multiple Queue architecture for process communication. 

**Request queue**: Endpoint puts the request along with an unique id (request id)

**Response queue**: Model worker puts the response along with the request id

**Async background response queue consumer**: A background async task will consume the response queue and put the result value that can be consumed by the endpoint. 

**TODOs**:
- [x] Update code for streaming and non-streaming flow
- [x] Update code for OpenAISpec
- [ ] Update tests

Fixes # (issue): Improves overall throughput 

## PR review

## Did you have fun?

### DIFF
--- a/src/litserve/server.py
+++ b/src/litserve/server.py
@@ -15,14 +15,11 @@ import asyncio
 import contextlib
 import copy
 import logging
-from collections import deque
 import pickle
 from contextlib import asynccontextmanager
 import inspect
-import threading
 import multiprocessing as mp
-from multiprocessing import Manager, Queue, Pipe
-from multiprocessing.connection import Connection
+from multiprocessing import Manager, Queue
 from queue import Empty
 import uvicorn
 import time
@@ -33,7 +30,6 @@ import uuid
 
 from fastapi import FastAPI, Depends, HTTPException, BackgroundTasks, Request, Response
 from fastapi.security import APIKeyHeader
-import sys
 
 from fastapi.responses import StreamingResponse
 from starlette.middleware.gzip import GZipMiddleware
@@ -42,7 +38,7 @@ from litserve import LitAPI
 from litserve.connector import _Connector
 from litserve.specs import OpenAISpec
 from litserve.specs.base import LitSpec
-from litserve.utils import wait_for_queue_timeout, LitAPIStatus, load_and_raise
+from litserve.utils import LitAPIStatus, load_and_raise
 
 logger = logging.getLogger(__name__)
 
@@ -334,7 +330,7 @@ def api_key_auth(x_api_key: str = Depends(APIKeyHeader(name="X-API-Key"))):
         )
 
 
-async def response_queue_to_buffer(response_queue: mp.Queue, buffer: Dict, stream:bool):
+async def response_queue_to_buffer(response_queue: mp.Queue, buffer: Dict, stream: bool):
     while True:
         try:
             uid, payload = response_queue.get_nowait()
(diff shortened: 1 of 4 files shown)


tests/test_simple.py
```diff
@@ -89,7 +89,7 @@ def encode_response(self, output) -> Response:
 
 
 @pytest.mark.asyncio()
-async def test_timeout():
+async def disable_test_timeout():
```
aniketmaurya, line 92: looking into ways to restore this. Since we removed the request_dict, we can't check if the request has been picked up from queue for processing. 

Also, there is no way to iterate MP Queue without removing items.

src/litserve/server.py
```diff
@@ -122,41 +121,58 @@ def run_single_loop(lit_api: LitAPI, lit_spec: LitSpec, request_queue: Queue, re
...
             batch_timeout,
         )
         if not batches:
             continue
 
         logger.debug(f"{len(batches)} batched requests received")
-        inputs, pipes = zip(*batches)
+        uids, timestamps, inputs = zip(*batches)
+
+        # Iterate in reverse order to safely remove items while iterating
```
lantiga, line 156: I would more simply make a copy of timestamps (they are not a lot), build a list of indices that need to be evicted, and then create a new list of uids, inputs, and timestamps accordingly

lantiga: But more in general, this is not the correct time to do this.
We need to do this in collate_requests, because batches should not contain timed out requests anyway.

At this point collate_requests needs to return
batches, timed_out_uids
instead of just batches, so this function can return the HTTP exception.

src/litserve/server.py
```diff
@@ -222,41 +244,57 @@ def run_streaming_loop(lit_api: LitAPI, lit_spec: LitSpec, request_queue: Queue,
...
-            request_buffer,
             max_batch_size,
             batch_timeout,
         )
         if not batches:
             continue
 
-        inputs, pipes = zip(*batches)
+        uids, timestamps, inputs = zip(*batches)
+        # Iterate in reverse order to safely remove items while iterating
```
lantiga, line 278: Same comment as above

src/litserve/server.py
```diff
@@ -344,10 +380,24 @@ def api_key_auth(x_api_key: str = Depends(APIKeyHeader(name="X-API-Key"))):
...
+    response_queue: mp.Queue, buffer: Dict[str, Tuple[asyncio.Queue, asyncio.Event]], stream: bool
+):
+    while True:
+        try:
+            uid, payload = response_queue.get_nowait()
+        except Empty:
+            await asyncio.sleep(0.0001)
+            continue
+
+        if stream:
```
lantiga, line 393: please condition `if stream` outside `while`, and have two different `while loops`

otherwise we keep on evaluating `if stream` but there's really no need to

src/litserve/specs/openai.py
```diff
@@ -311,35 +306,30 @@ async def options_chat_completions(self, request: Request):
...
-            read, write = self._server.new_pipe()
             request_el = request.model_copy()
             request_el.n = 1
-            self._server.request_buffer[uid] = (request_el, write)
-            self._server.request_queue.put(uid)
-            background_tasks.add_task(self._server.cleanup_request, self._server.request_buffer, uid)
-            pipes.append((read, write))
-
-        responses = await self.get_from_pipes(uids, pipes)
+            q = asyncio.Queue()
```
lantiga, line 316: I still think this might be overkill for this specific case. Let's just make a note to ourselves to try a simple `deque`.

lantiga (changes requested):
Timeout strategy needs to be tweaked for the batched cases.
Other than this, all looks good!

aniketmaurya:
thanks for reviewing @lantiga! applied the feedbacks!

lantiga (approved):
Stamped!

lantiga:
@aniketmaurya ready to merge, before doing so please run through the available studios for a final sanity check

PR #169, opened 2024-07-11
Fix: Support files > 1MB

andyland:
<details>

</details>

## What does this PR do?

This PR allows files that are larger than 1MB to be uploaded.

For files larger than 1MB uploaded with FastAPI, it creates a [SpooledTemporaryFile](https://docs.python.org/3/library/tempfile.html#tempfile.SpooledTemporaryFile) which is partially written to disk as [described here](https://fastapi.tiangolo.com/tutorial/request-files/#uploadfile).  This prevents pickling of the form by the multiprocess library and crashes the request.

This PR sets the limit to an extremely large number so the file is always kept entirely in memory.  This allows the file to be serialized correctly.  Effectively, this limits the max file size for LitServe to the amount of RAM available on the machine.

## PR review

## Did you have fun?

### DIFF
--- a/src/litserve/server.py
+++ b/src/litserve/server.py
@@ -19,6 +19,7 @@ import multiprocessing as mp
 import os
 import pickle
 import shutil
+import sys
 import time
 import uuid
 from contextlib import asynccontextmanager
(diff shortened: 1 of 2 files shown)


aniketmaurya (approved):
LGTM 🔥 . Thanks @andyland!

lantiga:
about this, we should have the ability to reject large uploads (max size configurable), or a ill intentioned consumer of the api could upload a huge file and crash the server

andyland:
@lantiga Yeah, this PR works around a technical limitation preventing us from supporting larger requests.  For that, I would recommend a different approach.  I'd attach a middleware to FastAPI to return a proper 413 HTTP code on some configurable size.

lantiga:
yes definitely, that’s also what I was thinking above
@aniketmaurya can you add an issue to track it?

aniketmaurya:
created an issue here #171

PR #174, opened 2024-07-17
Add max payload size middleware

### LINKED ISSUE #171: Prevent large payload attack
## 🚀 Feature

Enable users to configure the maximum payload size for requests. LitServe will reject requests larger than the configured payload size preventing the server from crash.

andyland:
<details>

</details>

## What does this PR do?

Fixes #171 .  Adds a configurable option for max payload size, that will inject a middleware that will return 413 (Payload Too Large) for requests too large

## PR review

## Did you have fun?

### DIFF
--- a/tests/test_form.py
+++ b/tests/test_form.py
@@ -33,17 +33,41 @@ class SimpleFileLitAPI(LitAPI):
 
 
 def test_multipart_form_data(tmp_path):
-    server = LitServer(SimpleFileLitAPI(), accelerator="cpu", devices=1, workers_per_device=1)
+    file_length = 1024 * 1024 * 100
+
+    server = LitServer(
+        SimpleFileLitAPI(), accelerator="cpu", devices=1, workers_per_device=1, max_payload_size=(file_length * 2)
+    )
 
     with TestClient(server.app) as client:
         file_path = f"{tmp_path}/big_file.txt"
-        file_length = 1024 * 1024 * 100
         with open(file_path, "wb") as f:
             f.write(bytearray([1] * file_length))
         with open(file_path, "rb") as f:
             file = {"input": f}
             response = client.post("/predict", files=file)
             assert response.json() == {"output": file_length**2}
+
+
+def test_file_too_big(tmp_path):
+    file_length = 1024 * 1024 * 100
+
+    server = LitServer(
+        SimpleFileLitAPI(), accelerator="cpu", devices=1, workers_per_device=1, max_payload_size=(file_length / 2)
+    )
+
+    with TestClient(server.app) as client:
+        file_path = f"{tmp_path}/big_file.txt"
+        with open(file_path, "wb") as f:
+            f.write(bytearray([1] * file_length))
+        with open(file_path, "rb") as f:
+            file = {"input": f}
+            response = client.post("/predict", files=file)
+            assert response.status_code == 413
+
+            # spoof content-length size
+            response = client.post("/predict", files=file, headers={"content-length": "1024"})
+            assert response.status_code == 413
 
 
 class SimpleFormLitAPI(LitAPI):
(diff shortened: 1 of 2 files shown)


tests/test_form.py
```diff
@@ -46,6 +49,24 @@ def test_multipart_form_data(tmp_path):
...
+    )
+
+    with TestClient(server.app) as client:
+        file_path = f"{tmp_path}/big_file.txt"
+        file_length = 1024 * 1024 * 100
+        with open(file_path, "wb") as f:
+            f.write(bytearray([1] * file_length))
+        with open(file_path, "rb") as f:
+            file = {"input": f}
+            response = client.post("/predict", files=file)
```
aniketmaurya, line 66: what happens if a bad actor modifies the content length:

```py
response = client.post("/predict", files=file, headers={"content-length": "1024"})
```

aniketmaurya (commented):
Looks good! added a question.

tests/test_form.py
andyland: Seems like this is open to spoofing.  This seems like an underlying problem with starlette that it does not cease reading when the content-length is reached.  I guess if the user is really worried about it, they should just authenticate their endpoint

bhimrazy: Hi @aniketmaurya @andyland  , Sorry for interrupting.
Do you think checking the actual body size might help ? 
Perhaps something like this:

```python
body = await request.body()
if len(body) > self.max_size:
    return ...
```

aniketmaurya: that's a good point @bhimrazy! We would need to iteratively read the request data before loading the full data in system memory and raise the MaxSizeError if it exceeds the provided value.

cc: @lantiga

lantiga: yes good point @bhimrazy 
it would add a synchronization point in the async chain, we should compare the effect of @bhimrazy's suggestion on the performance benchmarks we have and if it's negligible include it @aniketmaurya can you make that check?

andyland: @bhimrazy by reading the body, you are loading it into memory which would defeat the point of the check.

cc @lantiga @aniketmaurya

andyland: @aniketmaurya you would need to reimplement this function to read it iteratively: https://github.com/encode/starlette/blob/6f863b0d3b8e8f18d5df9e8cd2514f7085b874e1/starlette/requests.py#L221

andyland: Or we could just switch to this when its merged upstream: https://github.com/encode/starlette/pull/2328/files

aniketmaurya: yes @andyland, overriding the `Request.body` method to check for size iteratively would be the right way to go because as mentioned it defeats the purpose of large data attacks. 

This [Starlette PR](https://github.com/encode/starlette/pull/2328/files) seem to be open for way too long so maybe we should just do it on our side for now.

aniketmaurya:
This middleware measures the size in without breaking async. @andyland would you mind if I commit to this PR directly?

```py
class MaxSizeMiddleware:
    def __init__(self, app, max_size):
        self.app = app
        self.max_size = max_size

    async def __call__(self, scope, receive, send):
        if scope["type"] != "http":
            await self.app(scope, receive, send)
            return

        body_size = 0

        async def receive_request_body_size():
            nonlocal body_size

            message = await receive()
            assert message["type"] == "http.request"

            body_size += len(message.get("body", b""))

            if not message.get("more_body", False):
                print(f"Size of request body was: {body_size} bytes")

            if body_size > self.max_size:
                raise HTTPException(
                    status_code=413,
                    detail={"error": f"Request body too large {body_size}"},
                )

            return message

        await self.app(scope, receive_request_body_size, send)

```

tests/test_form.py
andyland: ok updated @aniketmaurya

aniketmaurya (approved):
Thanks @andyland! Looks great 🚀

PR #190, opened 2024-08-02
cleanup: fix test naming convention

aniketmaurya:
<details>

</details>

## What does this PR do?

Fixes # (issue).

## PR review

## Did you have fun?

### DIFF
--- a/tests/test_batch.py
+++ b/tests/test_batch.py
@@ -12,20 +12,20 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import asyncio
-from unittest.mock import MagicMock, patch
-from asgi_lifespan import LifespanManager
 import time
-from litserve.server import run_batched_loop
 from queue import Queue
-import pytest
+from unittest.mock import MagicMock, patch
 
+import pytest
+import torch
+import torch.nn as nn
+from asgi_lifespan import LifespanManager
 from fastapi import Request, Response
 from httpx import AsyncClient
 
 from litserve import LitAPI, LitServer
+from litserve.server import run_batched_loop
 from tests.conftest import wrap_litserve_start
-import torch
-import torch.nn as nn
 
 
 class Linear(nn.Module):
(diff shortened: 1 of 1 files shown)


awaelchli (approved):

tests/test_batch.py
```diff
@@ -63,7 +64,7 @@ def encode_response(self, output) -> Response:
         return {"output": float(output)}
 
 
-class SimpleLitAPI2(LitAPI):
+class SimpleTorchAPI(LitAPI):
```
Andrei-Aksionov, line 67: Maybe I'm missing a point, but since it's a class inherited from `LitAPI` it should not be called `SimpleTorchAPI`.
As I understand the class above is a batch version of this class, then maybe this one should be `SimpleLitAPI` and the one above should be `class SimpleBatchLitAPI(SimpleLitAPI)`?

aniketmaurya: valid point @Andrei-Aksionov, the reason I did't use the suggested names is because we already have `LitServe.examples.SimpleLitAPI` and `SimpleBatchLitAPI` for a different purpose.

tests/test_batch.py
```diff
@@ -22,9 +22,10 @@
 from asgi_lifespan import LifespanManager
 from fastapi import Request, Response
 from httpx import AsyncClient
+
 from litserve import LitAPI, LitServer
-from tests.conftest import wrap_litserve_start
 from litserve.server import run_batched_loop
+from tests.conftest import wrap_litserve_start
```
Borda, line 28: I would rather reserve `conftest` for fixtures and this (seems to be) general functionality move to another utils module

Borda (approved):

tests/test_batch.py
aniketmaurya: great suggestion @Borda! I will create an issue for a follow up PR. Or in case if you're interested in doing that please feel free go ahead :D

PR #192, opened 2024-08-05
properly shutdown litserve workers

aniketmaurya:
<details>

</details>

## What does this PR do?

Fixes # (issue).

## PR review

## Did you have fun?

### DIFF
--- a/src/litserve/server.py
+++ b/src/litserve/server.py
@@ -546,6 +546,13 @@ class LitServer:
     async def lifespan(self, app: FastAPI):
         loop = asyncio.get_running_loop()
 
+        if not hasattr(self, "response_queues") or not self.response_queues:
+            raise RuntimeError(
+                "Response queues have not been initialized. "
+                "Please make sure to call the 'launch_inference_worker' method of "
+                "the LitServer class to initialize the response queues."
+            )
+
         response_queue = self.response_queues[app.response_queue_id]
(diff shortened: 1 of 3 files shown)


awaelchli (approved):
A test with a side-effect that raises an exception and asserts terminate is called would be useful here

src/litserve/server.py
```diff
@@ -717,15 +717,16 @@ def run(
...
+        try:
+            servers = self._start_server(port, num_uvicorn_servers, log_level, sockets, uvicorn_worker_type)
+            for s in servers:
+                s.join()
+        finally:
+            print("Shutting down LitServe")
+            for w in litserve_workers:
+                w.terminate()
+                w.join()
+            manager.shutdown()
```
Borda, line 729: why the finally is needed?

aniketmaurya: so that we always make sure that the LitServer workers are terminated after terminating the uvicorn server

aniketmaurya:
thanks for the feedback @awaelchli! added a test to assert shutdown is called.

src/litserve/server.py
Borda: so in fact, you do?
```py
try:
    servers = self._start_server(port, num_uvicorn_servers, log_level, sockets, uvicorn_worker_type)
    for s in servers:
        s.join()
    print("Shutting down LitServe")
    for w in litserve_workers:
        w.terminate()
        w.join()
    manager.shutdown()
except:
    pass
```

Issue #195, opened 2024-08-06
move `wrap_litserve_start` to utils


aniketmaurya:
> I would rather reserve `conftest` for fixtures and this (seems to be) general functionality move to another utils module

_Originally posted by @Borda in https://github.com/Lightning-AI/LitServe/pull/190#discussion_r1705957686_

ankitsharma07:
hey i would like to take this up..

aniketmaurya:
feel free to go ahead @ankitsharma07 🚀

ankitsharma07:
> feel free to go ahead @ankitsharma07 🚀 

thanks.

aniketmaurya:
@ankitsharma07 let me know if you need any help.

ankitsharma07:
@aniketmaurya you can close this one

PR #201, opened 2024-08-10
moved wrap_litserve_start to utils

ankitsharma07:
#195 move wrap_litserve_start to utils

i moved the function inside src > litserve > utils and then reimported them inside all the tests files.

### DIFF
--- a/src/litserve/utils.py
+++ b/src/litserve/utils.py
@@ -17,11 +17,16 @@ import pickle
 import uuid
 from typing import Coroutine, Optional
 from contextlib import contextmanager
-from litserve.server import LitServer
+from typing import TYPE_CHECKING  
+
+
 
 from fastapi import HTTPException
 from starlette.middleware.base import BaseHTTPMiddleware
+
 from starlette.types import ASGIApp, Message, Receive, Scope, Send
+if TYPE_CHECKING:  
+    from litserve.server import LitServer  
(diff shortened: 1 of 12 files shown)


ankitsharma07:
@aniketmaurya hi, some checks are failing but i have moved the functions and its usages, can you check?

src/litserve/utils.py
```diff
@@ -16,6 +16,8 @@
 import pickle
 import uuid
 from typing import Coroutine, Optional
+from contextlib import contextmanager
+from litserve.server import LitServer
```
aniketmaurya, line 20:
```python
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from litserve.server import LitServer
```

src/litserve/utils.py
```diff
@@ -69,6 +71,18 @@ async def azip(*async_iterables):
         yield tuple(results)
 
 
+@contextmanager
+def wrap_litserve_start(server: LitServer):
```
aniketmaurya, line 75:
```suggestion
def wrap_litserve_start(server: "LitServer"):
```

aniketmaurya (commented):
Looks great @ankitsharma07! Added a suggestion to fix the failing tests.

aniketmaurya (approved):

PR #208, opened 2024-08-21
Feat: Evict requests if the client has disconnected

### LINKED ISSUE #165: Evict requests if the client has disconnected
## 🚀 Feature

If a client sends a request and is disconnected, LitServe still processes the request which is a waste of computation. We need to evict the request if the client is disconnected.

bhimrazy:
<details>

</details>

## What does this PR do?
Fixes #165.

This PR addresses the situation when client requests are disconnected before completion. It tracks canceled requests and stops ongoing/running tasks, thereby saving computational resources.

#### Approach
The solution involves checking whether the client has disconnected during both predict and stream predict operations. 
If a disconnection is detected, the system adds the corresponding request ID to the `request_evicted_status` multiprocessing dictionary. 

This dictionary is then monitored by the running loops (both streaming and non-streaming) in worker process. If the worker loop detects that the current running request ID is present in the `request_evicted_status`, it immediately terminates the ongoing task associated with that request.

#### Potential Impacts 
This approach might impact performance due to the additional overhead of monitoring and terminating tasks, which could introduce minor delays in processing. [Benchmarking](https://lightning.ai/lightning-ai/studios/litserve-performance-tuning-and-benchmarking)
image

---
#### TODO
- [x] Handle client request disconnection in streaming mode
- [ ]  Handle client request disconnection in non-streaming mode (In progress)
     - Investigating methods to terminate tasks from the worker process (`run_single_loop`).
- [ ] Add/Improve tests (In progress)
     - Exploring ways to verify task termination from the worker process
- [ ] Handle client disconnection for batched requests
     
> Any help or guidance on this PR would be greatly appreciated 🙏. Thank you! 😊

## PR review

## Did you have fun?

### DIFF
--- a/src/litserve/server.py
+++ b/src/litserve/server.py
@@ -665,7 +665,7 @@ class LitServer:
             response, status = None, None
             try:
                 while not task.done():
-                    if await request.is_disconnected():
+                    if hasattr(request, "is_disconnected") and await request.is_disconnected():
                         task.cancel()
                         break
                 response, status = await task
--- a/src/litserve/server.py
+++ b/src/litserve/server.py
@@ -580,6 +580,7 @@ class LitServer:
         return [f"{accelerator}:{device}"]
 
     async def data_streamer(self, q: deque, data_available: asyncio.Event, send_status: bool = False):
+        uid = None
         while True:
             try:
                 await data_available.wait()
@@ -602,8 +603,9 @@ class LitServer:
                         yield data
                 data_available.clear()
             except asyncio.CancelledError:
-                self.request_evicted_status[uid] = True
-                logger.error("Request evicted for the uid=%s", uid)
+                if uid is not None:
+                    self.request_evicted_status[uid] = True
+                    logger.error("Request evicted for the uid=%s", uid)
                 break
             except Exception as e:
                 # Handle other exceptions that might occur
--- a/src/litserve/server.py
+++ b/src/litserve/server.py
@@ -665,7 +665,7 @@ class LitServer:
             response, status = None, None
             try:
                 while not task.done():
-                    await asyncio.sleep(0.5)
+                    await asyncio.sleep(0.1)
                     if await request.is_disconnected():
                         task.cancel()
                         break
--- a/src/litserve/server.py
+++ b/src/litserve/server.py
@@ -665,7 +665,7 @@ class LitServer:
             response, status = None, None
             try:
                 while not task.done():
-                    await asyncio.sleep(0.1)
+                    await asyncio.sleep(0.5)
                     if await request.is_disconnected():
                         task.cancel()
                         break
--- a/src/litserve/server.py
+++ b/src/litserve/server.py
@@ -670,15 +670,23 @@ class LitServer:
             response_task = asyncio.create_task(wait_for_response())
             disconnection_task = asyncio.create_task(check_disconnection())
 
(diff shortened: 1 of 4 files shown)


aniketmaurya:
looking good @bhimrazy so far! We might have to run some benchmarks to verify that we don't lose performance because of multiprocessing synchronization. But really good approach 😄

bhimrazy:
> looking good @bhimrazy so far! We might have to run some benchmarks to verify that we don't lose performance because of multiprocessing synchronization. But really good approach 😄

Thanks, @aniketmaurya! for the feedback. 
I'm currently exploring ways to test this feature and will keep you updated on the progress.

aniketmaurya:
@bhimrazy I'd suggest to go ahead with this technique and maybe implement is for single non batched loop then we can run some tests. and if everything goes well then we can implement it for other loops too.

bhimrazy:
Sure, that sounds great!

tests/test_lit_server.py
```diff
@@ -120,6 +121,25 @@ async def test_stream(simple_stream_api):
             ), "Server returns input prompt and generated output which didn't match."
 
 
+@pytest.mark.asyncio()
+async def test_stream_client_disconnection(simple_stream_api, caplog):
```
bhimrazy, line 125: might need to slightly change the approach to testing. The task seems to get completed before it is cancelled.

bhimrazy:
Hi @aniketmaurya, 
I am looking for abit of guidance for cancelling the task which already started in `run_single_loop` (non streaming case).
> specifically this part from `run_single_loop` 
```python
  x = _inject_context(
      context,
      lit_api.decode_request,
      x_enc,
  )
  y = _inject_context(
      context,
      lit_api.predict,
      x,
  )
  y_enc = _inject_context(
      context,
      lit_api.encode_response,
      y,
  )
```


Also, I was testing with the following approach but wanted to get your opinion on whether this is a good way to handle it:
```python
def process_steps(...):
    # Handle all steps: decode, predict, encode
    return ...

task = asyncio.create_task(process_steps(...))

while not task.done():
    # Check if the request was cancelled from the predict function
    if cancelled:
        task.cancel()
        print("Task canceled")
        return None
    await asyncio.sleep(...)  # Check every ...ms

y_enc = await task
```
I'd really appreciate your insights. Thank you!

aniketmaurya:
sure @bhimrazy, I will ping you on discord

src/litserve/server.py
```diff
@@ -625,8 +657,21 @@ async def predict(request: self.request_type, background_tasks: BackgroundTasks)
 
             self.request_queue.put_nowait((response_queue_id, uid, time.monotonic(), payload))
 
-            await event.wait()
-            response, status = self.response_buffer.pop(uid)
+            async def wait_for_response():
```
bhimrazy, line 660: @aniketmaurya , This approach for non streaming case seems to impact the benchmark a lot. 
In this portion, it is just checking if the client is disconnected and the actual task is yet to be cancelled from running loop.

bhimrazy: Seems to got improved with abit of refactoring the approach 
image

williamFalcon:
@bhimrazy please add a proper PR description. LitServe is live now, so it needs to follow Lightning AI's guidelines for production-ready OSS software.

- clear PR descriptions
- tests
- documentation

thanks!

bhimrazy:
Closing this PR. 

Due to the complexity involved, the streaming and non-streaming cases will be handled separately in new PRs (in more better and cleaner way). 

You can find the streaming case PR here: #223.

PR #220, opened 2024-08-25
Enable batch-unbatch by default

aniketmaurya:
<details>

</details>

## What does this PR do?

This PR makes `LitAPI.batch` and `LitAPI.unbatch` enable by default where it returns the input and output as list if not implemented.

## PR review

## Did you have fun?

### DIFF
--- a/src/litserve/api.py
+++ b/src/litserve/api.py
@@ -85,11 +85,7 @@ class LitAPI(ABC):
 
             return numpy.stack(inputs)
 
-        if self.stream:
-            message = no_batch_unbatch_message_stream(self, inputs)
-        else:
-            message = no_batch_unbatch_message_no_stream(self, inputs)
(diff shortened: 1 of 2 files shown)


lantiga (approved):
Looks great.

Caveat: this
https://github.com/Lightning-AI/LitServe/blob/144476b758e9f2b14c4aa90d4ed1361e32fd1e47/src/litserve/api.py#L24
and
https://github.com/Lightning-AI/LitServe/blob/144476b758e9f2b14c4aa90d4ed1361e32fd1e47/src/litserve/api.py#L40
are now both unused and the error message is inaccurate. We could just remove.

At first I thought we could catch an exception but the error from incorrect batching would show up in predict and encode_response, so I don't think we can catch it without producing confusing false positives.

PR #223, opened 2024-08-26
Feat/evict req on client disconnect streaming case

### LINKED ISSUE #165: Evict requests if the client has disconnected

bhimrazy:
<details>

</details>

## How does this PR impact the user?
As a user, I want the server to stop processing requests that disconnect from the client before finishing. This PR focuses on tracking disconnected requests (specifically for non-batched streaming mode) and stops those running tasks, saving resources and freeing up space for handling other requests.

## What does this PR do?
Partially fixes #165.
- Handles client request disconnection in streaming mode (non-batch).

## PR review

## Did you have fun?


src/litserve/server.py
```diff
@@ -568,26 +580,33 @@ def device_identifiers(self, accelerator, device):
...
-                    yield data
-            data_available.clear()
+                    else:
+                        yield data
+                data_available.clear()
+            except asyncio.CancelledError:
+                if uid is not None:
+                    self.request_evicted_status[uid] = True
+                    logger.error("Request evicted for the uid=%s", uid)
+                break
```
aniketmaurya, line 609:
```suggestion
                return
```

aniketmaurya (approved):
Great work @bhimrazy! looking good.

aniketmaurya (commented):
I will run some perf test for streaming before merging this PR!

bhimrazy:
Hi @aniketmaurya,
I've tried with some modifications, opting to check at specific intervals rather than on each output. 

However, if this PR is impacting performance, I'm completely okay with closing it. 
We can explore alternative solutions that may be more effective.

lantiga:
Hey @bhimrazy thanks for the patience, we're going to benchmark soon!

src/litserve/loops.py
```diff
@@ -279,7 +280,11 @@ def run_streaming_loop(
                 lit_api.encode_response,
                 y_gen,
             )
-            for y_enc in y_enc_gen:
+            check_interval = 50
+            for index, y_enc in enumerate(y_enc_gen):
+                if index % check_interval == 0 and request_evicted_status.get(uid):
+                    request_evicted_status.pop(uid)
+                    break
```
bhimrazy, line 287: Checking the `request_evicted_status` for each token appears to have a significant impact, reducing performance from 3600 to around 3100. However, it may not be necessary to perform this check on every token.

While adding a check interval helps reduce the overhead and brings the performance closer to that of the main branch, but it still doesn't feel like an ideal solution.

aniketmaurya: yeah, and in case when the time-to-first-token is large but rest of the token stream speed is fast, it doesn't help much.

aniketmaurya: this is just single worker. with multiple workers it might impact even more.

lantiga: I think the overall design is correct, we are just way too aggressive checking the distributed dict and we get into contention problems.

One alternative that could reduce contention is getting a snapshot of the disconnected dictionary in every worker loop: so not use a managed dict but a shared value that the server publishes and that gets read as a whole by each worker periodically (every N seconds - we don't need a thread, we just check the time at every loop). This way every worker has a semi-up to date local dictionary that it can check as often as we want.

Having semi-up to date info on who disconnected every N seconds is totally fine, we don't need to react immediately.

This design also helps with ignoring items in the queue that come from clients that have been disconnected. For those we necessarily have to check at every request. If the local dictionary is not up to date we'll run some requests for nothing, but that's ok. One caveat is making sure the responses don't accumulate in the response dictionary on the webserver process, in this case (let's remember about this).

bhimrazy: Thank you, @lantiga, for the valuable insights. This approach seems promising. I'll take some time to study the concept and work on the implementation shortly.

aniketmaurya:
Hi @bhimrazy, closing this PR for now. At this point, feel free to reopen or create a new PR when you get back to this.

PR #244, opened 2024-08-29
test LitAPI decode-encode methods

aniketmaurya:
<details>

</details>

Describe (in plain English, not technical Jargon) how this improves the user experience. If you can't tie it back to a real tangible, user goal or describe it in plain english, it's a hint that this is likely not needed and is probably an "engineering nit". 

### GOOD:

### BAD:
"This PR enables GPUs". 

---- 

## What does this PR do?

Fixes # (issue).

Ensure that the decode_request and encode_response methods are thoroughly tested across various scenarios and inputs to enhance robustness.

**Increases the test coverage by 2%**.

```
@@         Coverage Diff         @@
##           main   #244   +/-   ##
===================================
+ Coverage    82%    84%   +2%     
===================================
  Files        13     13           
  Lines      1055   1054    -1     
===================================
+ Hits        861    881   +20     
+ Misses      194    173   -21     
```

## PR review

## Did you have fun?

### DIFF
--- a/src/litserve/specs/openai.py
+++ b/src/litserve/specs/openai.py
@@ -315,7 +315,7 @@ class OpenAISpec(LitSpec):
             message = output[-1]
         else:
             error = (
-                "Malformed output from LitAPI.predict: expected"
+                "Malformed output from LitAPI.predict: expected "
                 f"string or {{'role': '...', 'content': '...'}}, got '{output}'."
             )
             logger.error(error)
--- a/src/litserve/specs/openai.py
+++ b/src/litserve/specs/openai.py
@@ -308,6 +308,9 @@ class OpenAISpec(LitSpec):
         logger.debug(output)
         if isinstance(output, str):
             message = {"role": "assistant", "content": output}
+
+        elif self.validate_chat_message(output):
+            message = output
         elif isinstance(output, dict) and "content" in output:
             message = output.copy()
             message.update(role="assistant")
--- a/tests/test_litapi.py
+++ b/tests/test_litapi.py
@@ -102,13 +102,13 @@ def test_decode_request():
         api.decode_request({"input": "Hello"})
 
 
-def test_encode_response():
-    # Without spec
+def test_encode_response_without_spec():
     response = 4.0
     api = ls.examples.SimpleLitAPI()
     assert api.encode_response(response) == {"output": 4.0}, 'Encode response returns encoded output {"output": 4.0}'
 
-    # case: Use OpenAISpec implementation
+
+def test_encode_response_with_openai_spec():
     api = ls.examples.TestAPI()
     api._sanitize(max_batch_size=1, spec=ls.OpenAISpec())
     response = "This is a LLM generated text".split()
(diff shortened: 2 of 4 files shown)


src/litserve/specs/openai.py
```diff
@@ -311,16 +311,14 @@ def _encode_response(self, output: Union[Dict[str, str], List[Dict[str, str]]])
         elif isinstance(output, dict) and "content" in output:
             message = output.copy()
             message.update(role="assistant")
-        elif self.validate_chat_message(output):
-            message = output
```
Andrei-Aksionov, line 315: Why did you remove it?

I see that the check above partially matches `validate_chat_message`, perhaps that why?
But in the check above you overwrite `role` anyway, even if it was already in the dict.

So maybe keep it, but move above  `elif isinstance(output, dict) and "content" in output:`

Andrei-Aksionov (approved):

tests/test_litapi.py
```diff
@@ -79,3 +95,90 @@ def test_batch_unbatch_stream():
...
+    api._sanitize(max_batch_size=1, spec=ls.OpenAISpec())
+    request = ChatCompletionRequest(messages=[{"role": "system", "content": "Hello"}])
+    decoded_request = api.decode_request(request)
+    assert decoded_request[0]["content"] == "Hello", "Decode request should return the input message"
+
+
+def test_decode_request_with_openai_spec_wrong_request():
+    api = ls.examples.TestAPI()
+    api._sanitize(max_batch_size=1, spec=ls.OpenAISpec())
+    with pytest.raises(AttributeError, match="object has no attribute 'messages'"):
```
lantiga, line 117: let's extend the failure modes here

tests/test_litapi.py
```diff
@@ -79,3 +95,90 @@ def test_batch_unbatch_stream():
...
+        api.decode_request({"input": "Hello"})
+
+
+def test_encode_response():
+    response = 4.0
+    api = ls.examples.SimpleLitAPI()
+    assert api.encode_response(response) == {"output": 4.0}, 'Encode response returns encoded output {"output": 4.0}'
```
lantiga, line 124: let's extend encoding with more inputs types

tests/test_litapi.py
```diff
@@ -79,3 +95,90 @@ def test_batch_unbatch_stream():
...
+    generated_tokens = []
+    for output in api.encode_response(response):
+        generated_tokens.append(output["content"])
+    assert generated_tokens == response, f"Encode response should return the generated tokens {response}"
+
+
+def test_encode_response_with_openai_spec_invalid_input():
+    api = ls.examples.TestAPI()
+    api._sanitize(max_batch_size=1, spec=ls.OpenAISpec())
+    response = 10
```
lantiga, line 172: let's extend failure modes

lantiga (commented):
Looks good, but we should extend failure modes

tests/test_litapi.py
aniketmaurya: It is implemented by SimpleLitAPI so for more input types, we would be testing the SimpleLitAPI class then

tests/test_litapi.py
aniketmaurya: default decode_request is Pydantic typed so it handles anything other than a ChatCompletionRequest object with the same AttributeError.

Issue #146, opened 2024-06-19
API monitoring metrics


aniketmaurya:
## 🚀 Feature

Provide monitoring metrics: latency, decode-predict-encode time.

aniketmaurya:
thanks for the feedback @carlthome! I would love to know more about your use case with the monitoring metrics.

vrdn-23:
@aniketmaurya I think for any production use case we would want metrics that are similar to what we see in [TEI](https://github.com/huggingface/text-embeddings-inference) and [vllm](https://docs.vllm.ai/en/latest/serving/metrics.html#)

AdolfoVillalobos:
Hello @aniketmaurya!

I took my chances on having a first stab at this issue in PR #262 

I left some things open-ended because I wanted to get some feedback from the team. After some clarification, I will add better test support and so on.

Also, I only registered simple metrics for the prediction elapsed time. The idea is to clarify the setup:

> @aniketmaurya I think for any production use case we would want metrics that are similar to what we see in [TEI](https://github.com/huggingface/text-embeddings-inference) and [vllm](https://docs.vllm.ai/en/latest/serving/metrics.html#)


Then, as the links showcase, we can extend the coverage of the metric collection to a broader set of metrics, including tokens per unit of time and so on.

As a side note, adding a benchmark routine to the CI might be a good idea. 

Cheers!

aniketmaurya:
This can now be done using the Loggger and callbacks API - https://lightning.ai/docs/litserve/features/logger

PR #262, opened 2024-09-02
First attempt at monitoring metrics using prometheus_client

AdolfoVillalobos:
<details>

</details>

Describe (in plain English, not technical Jargon) how this improves the user experience. If you can't tie it back to a real tangible, user goal or describe it in plain english, it's a hint that this is likely not needed and is probably an "engineering nit". 

### GOOD:

### BAD:
"This PR enables GPUs". 

---- 

## What does this PR do?

First attempt at solving #146 

## PR review

This implementation introduces basic monitoring metrics using prometheus_client to collect API metrics. The goal is to provide a foundation for further enhancements and discussions.

Implementing monitoring involves some decision-making and tradeoffs, so I start by:
1.  proposing changes. 
2. Highlight some limitations that I didn't want to solve right away before getting feedback
3. Summarize the key issues to discuss.

I hope this PR proves to be a useful ground for discussion :)

#### 1. Prometheus Integration:

1. Utilizes prometheus_client to collect metrics.
2. Integrates seamlessly with FastAPI, automatically creating a /metrics endpoint.
3. Compatible with multithreading environments.

#### 2. Implementation Approach:
1. The current approach involves some repetition. A more elegant solution would be to use decorators for methods like predict, encode, and decode.
2. Care must be taken with callback annotations and signatures to maintain API flexibility, especially in a multithreading context.
#### 3. Memory Management:

Given the multithreading environment, memory management for the Prometheus client may be necessary.
This is an area that requires further discussion to determine if it's something the team wants to address now.

#### 1. Metric Scope:
1. Only prediction elapsed time is measured currently.
1. Generalization to other metrics and custom metrics is straightforward but not yet implemented.
#### 2. Code Repetition:
1. The current implementation is somewhat repetitive.
1. Future work should focus on reducing this through the use of decorators.
#### 3. Memory Management:
1. Potential memory management issues with Prometheus in a multithreading environment need to be considered.

### Next Steps (Summary)
1. Discuss and decide on the best approach for extending metrics. Should we use prometheus?
1. Evaluate the use of decorators to reduce code repetition. This might involve some refactoring.
1. Consider the implications of memory management for Prometheus in our multithreading setup. Maybe a simpler solution would be best?

## Did you have fun?

Yess. I've been playing with this issue for a couple of hours and I think I have a better grasp of the project.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -130,3 +130,5 @@ venv.bak/
 lightning_logs/
 MNIST
 .DS_Store
+
+metrics/
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,4 @@
 fastapi >=0.100
 httpx
 uvicorn >=0.29.0
+prometheus-client >=0.20.0
--- a/src/litserve/loops.py
+++ b/src/litserve/loops.py
@@ -33,7 +33,7 @@ from prometheus_client import Summary
 
 mp.allow_connection_pickling()
 
-PREDICT_METRIC = Summary('PREDICT_METRIC', 'A metric to summarize the predict execution times')
+PREDICT_METRIC = Summary("PREDICT_METRIC", "A metric to summarize the predict execution times")
 
 try:
     import uvloop
--- a/src/litserve/loops.py
+++ b/src/litserve/loops.py
@@ -28,7 +28,12 @@ from litserve import LitAPI
 from litserve.specs.base import LitSpec
 from litserve.utils import LitAPIStatus
 
+from prometheus_client import Summary
+
+
 mp.allow_connection_pickling()
+
+PREDICT_METRIC = Summary('PREDICT_METRIC', 'A metric to summarize the predict execution times')
 
 try:
     import uvloop
@@ -119,11 +124,13 @@ def run_single_loop(lit_api: LitAPI, lit_spec: LitSpec, request_queue: Queue, re
                 lit_api.decode_request,
                 x_enc,
             )
+            start_time = time.monotonic()
             y = _inject_context(
                 context,
                 lit_api.predict,
                 x,
             )
+            PREDICT_METRIC.observe(time.monotonic() - start_time)
             y_enc = _inject_context(
                 context,
                 lit_api.encode_response,
@@ -183,8 +190,10 @@ def run_batched_loop(
                 for input, context in zip(inputs, contexts)
             ]
             x = lit_api.batch(x)
+            start_time = time.monotonic()
             y = _inject_context(contexts, lit_api.predict, x)
             outputs = lit_api.unbatch(y)
+            PREDICT_METRIC.observe(time.monotonic() - start_time)
             for response_queue_id, y, uid, context in zip(response_queue_ids, outputs, uids, contexts):
                 y_enc = _inject_context(context, lit_api.encode_response, y)
 
@@ -228,11 +237,14 @@ def run_streaming_loop(lit_api: LitAPI, lit_spec: LitSpec, request_queue: Queue,
                 lit_api.decode_request,
                 x_enc,
             )
+
+            start_time = time.monotonic()
             y_gen = _inject_context(
                 context,
                 lit_api.predict,
                 x,
             )
+            PREDICT_METRIC.observe(time.monotonic() - start_time)
             y_enc_gen = _inject_context(
                 context,
                 lit_api.encode_response,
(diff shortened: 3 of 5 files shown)


aniketmaurya:
Hi @AdolfoVillalobos, great PR! I'm working on a callback system for LitServe which would act similar to middleware but gives you much more control over different parts of LitServe while serving a model.

I think this PR would be a great use case for the callback. I'm gonna create a PR for that soon and think that we can reorganize this PR to convert into a callback so that users can add or remove as required.

AdolfoVillalobos:
That sounds really cool! I found myself thinking about the same problem while I was investigating the system. Good to know!

I will update this PR as soon as you submit yours.

I will be looking for other opportunities to contribute to the project in the meantime!

AdolfoVillalobos:
@aniketmaurya I just realized the codecov/patch check on the CI is hanged. I can't cancel it. Should I worry? Cheers!

aniketmaurya:
There is no need to worry about the hanging CI, @AdolfoVillalobos. It's probably some issue with codecov service.

aniketmaurya:
hi @AdolfoVillalobos, callback system PR is merged. Let's do this!!

This PR can be refactored like the following: 

```python
from prometheus_client import Summary
from prometheus_client import multiprocess, CollectorRegistry, make_asgi_app

def make_metrics_app():
    generate_metrics_dir()
    registry = CollectorRegistry()
    multiprocess.MultiProcessCollector(registry)
    return make_asgi_app(registry=registry)

class PrometheusCallback(ls.Callback):
    def __init__(self):
        self.summary = Summary("PREDICT_METRIC", "A metric to summarize the predict execution times")

	def on_server_start(self, litserver):
        metrics_app = make_metrics_app()
		litserver.app.mount("/metrics", metrics_app)

    def on_before_predict(self, lit_api: "LitAPI"):
        t0 = time.perf_counter()
        self._start_time = t0

    def on_after_predict(self, lit_api: "LitAPI"):
        t1 = time.perf_counter()
        elapsed = t1 - self._start_time
        self.summary.observe(elapsed)
```

williamFalcon:
ok, this actually needs to be broken up into 2 things @aniketmaurya , a prometheus logger and a self.log api

## A telemetry callback
The callback for telemetry (and i would also add decode, encode request, etc...).  

```python
class SpeedTelemetry(ls.Callback):

    def on_before_predict(self, lit_api: "LitAPI"):
        t0 = time.perf_counter()
        self._start_time = t0

    def on_after_predict(self, lit_api: "LitAPI"):
        t1 = time.perf_counter()
        elapsed = t1 - self._start_time
        self.log('predict time', elapsed)
```

## A logger
All the setup stuff is really at the logger level.    

```python
Server(callbacks=[SpeedTelemetry()], loggers=[Prometheus(), ...])
```

AdolfoVillalobos:
> @williamFalcon created a [PR](https://github.com/Lightning-AI/LitServe/pull/284) to introduce the Logger API.

This is so cool. Thanks for the guidance. I will be taking a stab at this issue on the following days. Thanks!

vrdn-23:
Thanks for the work adding compatibility for this @aniketmaurya !
I just wanted to understand the end goal of this PR if possible. Now that the functionality for callbacks and loggers have been landed,  would the end user be in charge of implementing and exposing their own metrics or does LitServer plan to have a set of metrics built-in with the option to extend with custom metrics if needed?
If it is the latter, are there a set of metrics that are determined to be made available or is that list still being discussed? 
cc @williamFalcon

williamFalcon:
generally, metrics are very specific to each user/server/system. we will provide basic callbacks that people can add, but metrics can get expensive in terms of performance, so we want users to explicitly opt into the callbacks they use. 

but adding a metric you care about is now super straightforward with self.log(…) inside the litapi.

vrdn-23:
@williamFalcon sounds great! Thanks for the clarification! I guess in that case is there any estimated timeline for when the next release of LitServe is going to be cut?

vrdn-23:
I am currently trying to implement some custom metrics for our team and we would be interested in a metric that is similar to `queue_size` which provides the current number of requests waiting in the queue. But from what I can tell the `request_queue` property isn't exposed via the LitAPI interface, so is there some other way that I'm missing that can help calculate this?

aniketmaurya:
@vrdn-23 although you can use [queue.qsize()](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.Queue.qsize) method, it is approximate and not reliable. 

Also, a cleaner way to this would be to have a [middleware](https://lightning.ai/docs/litserve/features/callbacks-middleware#using-middleware-in-litserve) with a counter that increment the counter when it enters the request and decrement it after processing the request.

aniketmaurya:
You can use prometheus with LitServe now. Please follow the docs [here](https://lightning.ai/docs/litserve/features/logger#prometheus).

PR #269, opened 2024-09-03
add perf test CI 1/3

aniketmaurya:
<details>

</details>

No direct impact! This PR will help contributors make sure their PR doesn't cause any performance degradation. 

---- 

## What does this PR do?

Perf test CI to make sure PRs don't introduce any latency or performance degradation. 

This PR only focuses on non-streaming case. I'll create a follow up for streaming perf test.

## PR review

## Did you have fun?

### DIFF
--- a/tests/perf_test/bert/benchmark.py
+++ b/tests/perf_test/bert/benchmark.py
@@ -24,4 +24,5 @@ def get_average_throughput(num_requests=100, num_samples=10):
 
 
 if __name__ == "__main__":
-    get_average_throughput(100, num_samples=10)
+    rps = get_average_throughput(100, num_samples=10)
+    assert rps >= 350, f"Expected RPS to be greater than 350, got {rps}"
--- a/tests/perf_test/bert/benchmark.py
+++ b/tests/perf_test/bert/benchmark.py
@@ -6,6 +6,7 @@ from simple_benchmark import benchmark
 
 # Configuration
 SERVER_URL = "http://0.0.0.0:8000/predict"
+MAX_SPEED = 390  # Nvidia 3090
 
 session = requests.Session()
 
@@ -39,4 +40,4 @@ if __name__ == "__main__":
         time.sleep(10)
 
     rps = get_average_throughput(100, num_samples=10)
-    assert rps >= 350, f"Expected RPS to be greater than 350, got {rps}"
+    assert rps >= MAX_SPEED, f"Expected RPS to be greater than {MAX_SPEED}, got {rps}"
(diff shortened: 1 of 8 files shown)


tests/perf_test/bert/simple_benchmark.py
```diff
@@ -0,0 +1,152 @@
...
+
+
+# Configuration
+SERVER_URL = "http://0.0.0.0:8000/predict"
+
+session = requests.Session()
+
+executor = None
+
+phrases = [
```
Borda, line 26: let's create a data folder with these samples.

aniketmaurya: done

awaelchli (approved):

tests/perf_test/bert/benchmark.py
```diff
@@ -0,0 +1,43 @@
...
+
+if __name__ == "__main__":
+    for i in range(10):
+        try:
+            resp = requests.get("http://localhost:8000/health")
+            if resp.status_code == 200:
+                break
+        except requests.exceptions.ConnectionError as e:
+            logging.error(f"Error connecting to server: {e}")
+        time.sleep(10)
```
Andrei-Aksionov, line 40: In general, I recommend to make sleep smaller, but the number of iterations larger.
Right now it might not make a huge sense, but later on, when tests continue to accumulate, if you do the same in all tests the CI might be very slow.

Andrei-Aksionov (approved):

tests/perf_test/bert/benchmark.py
aniketmaurya: agree! the magnitude of time for server to start is in range of 10-30 secs so I added 10 secs of sleep here.

Borda, line 40: this would be probably better with standard pytest and mark it as [flaky for retry](https://pypi.org/project/pytest-retry/)
```py
@@pytest.mark.flaky(retries=10)
def test_...():
    resp = requests.get("http://localhost:8000/health")
    assert resp.status_code == 200:
```

tests/perf_test/bert/data.py
Borda, line 1: why not plain text (.txt) file?

tests/perf_test/bert/server.py
```diff
@@ -0,0 +1,64 @@
+import torch
```
Borda, line 1: lets add simple bite what it is about and with what client it shall be used?

Borda (approved):

tests/perf_test/bert/benchmark.py
aniketmaurya: added retry

tests/perf_test/bert/data.py
aniketmaurya: I used a Python list to keep the data directly accessible and easily manipulable within the code without the extra overhead of file I/O and parsing.

PR #279, opened 2024-09-17
minor LitServer code clean up

aniketmaurya:
<details>

</details>

No impact to users. Just code clean up

### GOOD:

### BAD:
"This PR enables GPUs". 

---- 

## What does this PR do?

Fixes # (issue).

## PR review

## Did you have fun?

### DIFF
--- a/src/litserve/server.py
+++ b/src/litserve/server.py
@@ -124,7 +124,7 @@ class LitServer:
 
         if middlewares is None:
             middlewares = []
-        if not isinstance(middlewares, list):
+        elif not isinstance(middlewares, list):
             _msg = (
                 "middlewares must be a list of tuples"
                 " where each tuple contains a middleware and its arguments. For example:\n"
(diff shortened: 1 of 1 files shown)


src/litserve/server.py
```diff
@@ -196,28 +196,22 @@ def __init__(
...
             server_copy = copy.copy(self)
             del server_copy.app
-            try:
-                spec.setup(server_copy)
-            except Exception as e:
-                raise e
+            spec.setup(server_copy)
 
         process_list = []
         for worker_id, device in enumerate(self.devices * self.workers_per_device):
```
bhimrazy, line 217: Can we use `self.inference_workers` instead of `self.devices * self.workers_per_device` since they refer to the same calculation?

bhimrazy (approved):

src/litserve/server.py
aniketmaurya: great catch! updated.

Andrei-Aksionov (approved):

PR #290, opened 2024-09-23
Feat: decode request in threadpool

### LINKED ISSUE #166: Map `decode_request` during dynamic batching using a threadpool
## 🚀 Feature

A default optimization that LitServe can provide users is to map the `decode_request` function in case of dynamic batching using a ThreadPool. This can be useful for cases like image loading which is IO based.

I did a quick test with a ResNet-152 model for image classification and observed the following performance gain with threadpool:
image

deependujha:
<details>

</details>

faster request decoding if it contains binary data (images, audios, etc).

### GOOD:

faster decoding in batched loop

### BAD:

Might be overhead in the case of simple requests

---- 

## What does this PR do?

Fixes #166 

Speed up reference [here](https://github.com/Lightning-AI/LitServe/issues/166)

## PR review

## Did you have fun?


src/litserve/loops.py
```diff
@@ -55,6 +56,16 @@ def _inject_context(context: Union[List[dict], dict], func, *args, **kwargs):
     return func(*args, **kwargs)
 
 
+def inject_context_in_threadpool(contexts: List[Union[List[dict], dict]], func: Callable, inputs: List[dict]):
+    with concurrent.futures.ThreadPoolExecutor(max_workers=min(len(inputs), len(contexts))) as executor:
```
aniketmaurya, line 60: let's create the threadpool at loop level instead of initializing every time.

deependujha:
Is this what was intended, or have I got something wrong?

aniketmaurya (commented):
Thank you for opening the PR @deependujha. Looks great so far 🚀 

We should provide control to users whether to enable decoding using a threadpool. It might come useful for cases when decode request is CPU intensive or not thread-safe. 

Possible API:

```python
class LitServer:
    def __init__(..., concurrent_decode:bool=True)
```

cc: @lantiga

williamFalcon:
@deependujha nice! love the "faster" claim haha. do you have a benchmark or something showing that it is indeed faster and by how much?

deependujha:
> @deependujha nice! love the "faster" claim haha. do you have a benchmark or something showing that it is indeed faster and by how much?

Hi @williamFalcon , I stated so on the basis of what was claimed in the original issue. Testing on 4 GPU machine does not reflect the same.

Throughput stays in the similar range (300-380), averaging at 330.

I'll try testing it with an audio model. If performance doesn't differ much, I'll close the PR.

src/litserve/loops.py
```diff
@@ -300,7 +323,11 @@ def run_batched_streaming_loop(
     max_batch_size: int,
     batch_timeout: float,
     callback_runner: CallbackRunner,
+    concurrent_decode: bool = True,
 ):
+    if concurrent_decode:
+        executor = concurrent.futures.ThreadPoolExecutor(max_workers=os.cpu_count())
```
aniketmaurya, line 329: We can limit number of threads to batch size
```suggestion
        executor = concurrent.futures.ThreadPoolExecutor(max_workers=min(max_batch_size, os.cpu_count()))
```

deependujha:
tough luck, tested with the audio model but no significant difference in performance. I'll go ahead and close the PR.

aniketmaurya:
@deependujha how does your `decode_request` logic look like? Does it do a CPU intensive operation or IO?

PR #300, opened 2024-09-25
Add warning for unexpected model output in batched prediction

### LINKED ISSUE #294: unexpected output for HF model with matching
### To Reproduce

without batching all works as expected
```
{'output': 'What is the capital of Greece?\n\nAthens.'}
```
but with batch, it returns just the first character
```
{'output': 'W'}
```

#### Code sample

```py
import torch
import litserve as ls
from transformers import AutoTokenizer, AutoModelForCausalLM


class JambaLitAPI(ls.LitAPI):

    def __init__(
        self,
        model_name: str = "ai21labs/AI21-Jamba-1.5-Mini",
        max_new_tokens: int = 100
    ):
        self.model_name = model_name
        self.max_new_tokens = max_new_tokens

    def setup(self, device):
        # Load the model and tokenizer from Hugging Face Hub
        # For example, using the `distilbert-base-uncased-finetuned-sst-2-english` model for sentiment analysis
        # You can replace the model name with any model from the Hugging Face Hub
        self.model = AutoModelForCausalLM.from_pretrained(
            self.model_name, torch_dtype=torch.bfloat16, device_map="auto", use_mamba_kernels=False
        ).eval()
        self.tokenizer = AutoTokenizer.from_pretrained(self.model_name, legacy=False)

        # Extract text from request
        # This assumes the request payload is of the form: {'input': 'Your input text here'}
        return request["input"]

    def predict(self, text):
        print(text)
        # Use the loaded pipeline to perform inference
        inputs = self.tokenizer(text, return_tensors='pt')
        input_ids = inputs.to(self.model.device)["input_ids"]
        print(input_ids)
        output_ids = self.model.generate(
            input_ids,
            max_new_tokens=self.max_new_tokens
        )[0]
        print(output_ids)
        text = self.tokenizer.decode(output_ids, skip_special_tokens=True)
        print(text)
        return text

        # Format the output from the model to send as a response
        # This example sends back the label and score of the prediction
        return {"output": output}


if __name__ == "__main__":
    # Create an instance of your API
    api = JambaLitAPI()
    # Start the server, specifying the port
    server = ls.LitServer(api, accelerator="cuda", devices=1, max_batch_size=4)
    # print("run the server...")
    server.run(port=8000)
```

### Environment
If you published a Studio with your bug report, we can automatically get this information. Otherwise, please describe:   

- PyTorch/Jax/Tensorflow Version (e.g., 1.0):
- OS (e.g., Linux):
- Python version:
- CUDA/cuDNN version:

grumpyp:
<details>

#294 
is this needed?

</details>

## What does this PR do?

Fixes #294 .

## PR review

## Did you have fun?

Yes, the Debugger is my new friend :D

## Question:

Do I need to add this warning to the docs?


src/litserve/loops.py
```diff
@@ -208,9 +208,14 @@ def run_batched_loop(
             callback_runner.trigger_event(EventTypes.BEFORE_PREDICT, lit_api=lit_api)
             y = _inject_context(contexts, lit_api.predict, x)
             callback_runner.trigger_event(EventTypes.BEFORE_PREDICT, lit_api=lit_api)
+            if isinstance(y, str):
+                logger.warning(
+                    "The 'predict' method returned a string instead of a list of predictions. "
+                    "When batching is enabled, 'predict' should return a list of predictions. "
+                    "To avoid unexpected behavior, ensure 'predict' returns a list of predictions, or implement 'LitAPI.unbatch' correctly."
+                )
```
aniketmaurya, line 216: The cleanest way to do this is to put it [here](https://github.com/Lightning-AI/LitServe/blob/e4d6b4aab221c3cc93daa49b413700e6d2d6a6ef/src/litserve/api.py#L66)

aniketmaurya (changes requested):

src/litserve/loops.py
grumpyp: I had something there in the beginning actually, will do it! Thanks.

tests/test_batch.py
```diff
@@ -148,6 +148,32 @@ def test_max_batch_size_warning():
...
+        y = api.predict(mock_input)
+        if isinstance(y, str):
+            import warnings
+
+            warnings.warn(
+                "The 'predict' method returned a string instead of a list of predictions. "
+                "When batching is enabled, 'predict' should return a list of predictions. "
+                "To avoid unexpected behavior, ensure 'predict' returns a list of predictions, "
+                "or implement 'LitAPI.unbatch' correctly."
+            )
```
aniketmaurya, line 174: why we need this code here? `api.predict` will already add this warning
```suggestion
```

src/litserve/api.py
```diff
@@ -63,6 +63,13 @@ def predict(self, x, **kwargs):
         pass
 
     def _unbatch_no_stream(self, output):
+        if isinstance(output, str):
+            warnings.warn(
+                "The 'predict' method returned a string instead of a list of predictions. "
+                "When batching is enabled, 'predict' should return a list of predictions. "
+                "To avoid unexpected behavior, ensure 'predict' returns a list of predictions, "
+                "or implement 'LitAPI.unbatch' correctly."
```
aniketmaurya, line 71: Predict should alway return outputs in correct format.
```suggestion
                "To avoid unexpected behavior, ensure 'predict' returns a list of predictions."
```

aniketmaurya (approved):
Thanks for updating the PR @grumpyp. Looks great! added minor comment and then we can merge.

PR #316, opened 2024-10-02
Add request queue to callbacks

vrdn-23:
<details>

- [x ] Was this discussed/agreed via a Github issue? (no need for typos and docs improvements)
- [x ] Did you read the [contributor guideline](https://github.com/Lightning-AI/pytorch-lightning/blob/main/.github/CONTRIBUTING.md), Pull Request section?
- [ x] Did you make sure to update the docs?
- [x ] Did you write any new necessary tests?

</details>

As a user, it would be beneficial to have the ability to track the progress of requests that re coming in through the LitServe workload. To that end, a good metric to decide on when the LitServe instance is being throttled is to measure the number of requests that are still waiting to be processed in the queue. In order to enable the easy calculation of this particular metric, this PR allows the `request_queue` parameter to also be enabled as part of the callback trigger.

If the changes are too extreme, it might also be worth pointing out that most queue metrics only need to be calculated as and when we start to decode a request and probably only needs to be added as part of the `on_before_decode_request` callback trigger.

## What does this PR do?

Fixes https://github.com/Lightning-AI/LitServe/pull/262#issuecomment-2384724504

## PR review

## Did you have fun?
Yes(?)

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -2,7 +2,7 @@
 
 # These owners will be the default owners for everything in the repo. Unless a later match takes precedence,
 # @global-owner1 and @global-owner2 will be requested for review when someone opens a pull request.
-* @lantiga @aniketmaurya @awaelchli @Andrei-Aksionov
+* @lantiga @aniketmaurya @ethanwharris @Andrei-Aksionov
 
 # CI/CD and configs
 /.github/       @borda
@@ -13,3 +13,4 @@
 /_requirements/ @borda @lantiga
 
 /README.md                           @williamfalcon @lantiga
+/requirements.txt                    @williamfalcon
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,3 @@
 fastapi >=0.100
 httpx
 uvicorn[standard] >=0.29.0
-jsonargparse[signatures]>=4.29.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,4 @@
 fastapi >=0.100
 httpx
 uvicorn[standard] >=0.29.0
+jsonargparse[signatures]>=4.29.0
--- a/setup.py
+++ b/setup.py
@@ -110,4 +110,9 @@ setup(
         "Programming Language :: Python :: 3.10",
         "Programming Language :: Python :: 3.11",
     ],
+    entry_points={
+        "console_scripts": [
+            "litserve=litserve.__main__:main",
+        ],
+    },
 )
--- a/src/litserve/__about__.py
+++ b/src/litserve/__about__.py
@@ -11,7 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-__version__ = "0.2.2"
+__version__ = "0.2.3"
 __author__ = "Lightning-AI et al."
 __author_email__ = "community@lightning.ai"
 __license__ = "Apache-2.0"
--- a/src/litserve/__main__.py
+++ b/src/litserve/__main__.py
@@ -11,19 +11,27 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-from argparse import ArgumentParser
+from argparse import ArgumentParser, ArgumentDefaultsHelpFormatter, RawTextHelpFormatter
 from litserve.docker_builder import dockerize
 
 
+class LitFormatter(ArgumentDefaultsHelpFormatter, RawTextHelpFormatter): ...
+
+
 def main():
-    parser = ArgumentParser(description="CLI for LitServe")
-    subparsers = parser.add_subparsers(dest="command", help="Sub-command help")
+    parser = ArgumentParser(description="CLI for LitServe", formatter_class=LitFormatter)
+    subparsers = parser.add_subparsers(
+        dest="command",
+        title="Commands",
+    )
 
     # dockerize sub-command
     dockerize_parser = subparsers.add_parser(
         "dockerize",
-        help=dockerize.__doc__,
-        description="Generate a Dockerfile for the given server code.",
+        help="Generate a Dockerfile for the given server code.",
+        description="Generate a Dockerfile for the given server code.\nExample usage:"
+        " litserve dockerize server.py --port 8000 --gpu",
+        formatter_class=LitFormatter,
     )
     dockerize_parser.add_argument(
         "server_filename",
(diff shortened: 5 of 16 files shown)


aniketmaurya:
Pasting the comments from [here](https://github.com/Lightning-AI/LitServe/pull/262#issuecomment-2387874434)

@vrdn-23 although you can use [queue.qsize()](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.Queue.qsize) method, it is approximate and not reliable.

A cleaner way to this would be to have a [middleware](https://lightning.ai/docs/litserve/features/callbacks-middleware#using-middleware-in-litserve) with a counter that increment the counter when it enters the request and decrement it after processing the request.

Another reason why it shouldn't be exposed is that - it is the core part to make the request flow from LitServer to LitAPI and if users update or access it then it can slow down or bring some weird behaviour.

vrdn-23:
> A cleaner way to this would be to have a middleware with a counter that increment the counter when it enters the request and decrement it after processing the request.

This does look promising. But I'm a little confused how the logger will have access to the value specified in the middleware? Is there a convenient way to access the middleware attributes from the lit_api variable that is made available to the callbacks?

aniketmaurya:
Middleware won't be able to connect to loggers. You can track the number of requests in queue separately using a [multiprocessing shared Value](https://github.com/Lightning-AI/LitServe) and use the same in Logger to log it.

aniketmaurya:
Closing this PR since we won't be adding `request_queue` to the callback hooks. If someone reading this thread also want to have this feature, please feel free to comment and we can evaluate this in future. Thank you ⚡

vrdn-23:
> Middleware won't be able to connect to loggers. You can track the number of requests in queue separately using a [multiprocessing shared Value](https://github.com/Lightning-AI/LitServe) and use the same in Logger to log it.

I think the wrong link may have been linked here @aniketmaurya. Either way I think I would like to make sure I document the problem properly so that it is understood why I think the current callback mechanism as it is will not be able to effectively solve it.

## Why do I think this will currently not work with LitServe?
The idea here is to be able to expose something that allows us to calculate the number of requests that are piled up in a queue waiting to be processed. In the current callback system, the very first callback [happens only after the requests](https://github.com/Lightning-AI/LitServe/blob/97f7192f8992f1c4e297f01971db6c94ae6cd326/src/litserve/loops.py#L195) from the queue have processed. This means that there is no plausible way for us to record how many messages are still remaining in the queue. Adding any sort of logging metrics within the callback system will give us an accurate count of how many requests have been processed or how many are concurrently going through each phase, but it will not expose the number of requests that are currently waiting in the queue. 

## Why do I want this feature implemented in LitServe?
The reason why I am keen to see this behavior implemented in LitServe is because I think the amount of requests queueing up is a very crisp and concise indicator of when we need to scale the number of model replicas up according to the current load. We currently host our models on Kubernetes deployments and we would like to use a more relaible metric that directly reflects the pressure the model server is under. (A good proxy for this would be downstream response latency which can tell us that a particular model request spent too much time in the queue but in this case it is unclear to us whether the extra time is due to the prediction time or due to the time the request waits in the queue.)

## Do I have a solution for this?
I agree with @aniketmaurya about the concerns he has raised with regarding exposing the request_queue. There are a couple of scenarios that I've considered that may help alleviate this problem (in order of preference):

- **Expose Prometheus metrics as a LitServe standard**: Prometheus is kind of the go-to standard when it comes to exposing metrics in most production environments which rely on Kuberntes. Having in-built metrics that are served during the server running, would be a minimum requirement for any user that hopes to gets this intergated in a cloud k8s environment (which is where most ML workloads run nowadays). 
I would like to volunteer my time and effort in helping LitServe expose a small set of standard metrics that I think will be useful from an operational standpoint so that it aligns with what is present in most mature model-serving frameworks. I understand @williamFalcon's concern with [performance metrics hindering performance](https://github.com/Lightning-AI/LitServe/pull/262#issuecomment-2376483558) but I do not expect that this is something that will be impactful enough to hinder any performance gains. (One way to actually verify this would be if we had some metrics available from running the server :) ) 
Providing these metrics as part of the LitServe standard will be a much more safer/cleaner approach as one could argue that I'm sure a user could inadvertently find ways to mess up the the exposed LitAPI object as well by updating or accessing it. 


- **Add a separate callback that runs before the batches are collated**: The other alternate option that I would like the maintainer team to consider which would be a much smaller and less involved fix would be to add a callback that is triggered before the requests are collated (something along the lines of `before_request_batching` or `after_request_batching`) and this can expose something innocuous like `queue.qsize()` as the callback args (which although I recognize as approximate and unreliable is still a better approximation than nothing at all). This should alleviate concerns of users having access to the `request_queue` object but still provide a way for users to keep track of messages in the queue if needed. This also maintains the argument structure for existing callbacks that only expose the `lit_api` argument and would act as a special callback just for queues if needed. 

I'm sorry if it appears that I'm being stubborn on this. I think this project is a phenomenal piece of work and I truly believe that LitServe has the potential to fill the gap that exists in today's state of model deployment infrastructure where we need to host models that are slightly non-conventioal in either set-up or not fully adopted by the mainstream ML field. 

I respect the importance that the maintainers have on not over-complicating the code base but I do think that there are certain features from an operational standpoint that make working with open-source libraries much easier to adopt in a production environment and the cost of maintaining a lean library should not stand in the way of providing a flexible and robust framework.

If the decision stands the same, I will respect that but I do think it is worth re-considering at-least. Looking forward to hearing your thoughts on this!

lantiga:
I think being able to know the approximate size of the queue and track it over time is important. We just need to decide where to do it best.

I agree with @aniketmaurya that exposing the queue is too extreme.
Since this metric is useful under heavy load we need to provide the value proactively in a way that has fewer chances to introduce additional latency.

The middleware route is a good one, we could add a hook similar to on_server_start (that is, triggered on the webserver and not in the workers) that passes the number of active requests (incoming vs served). Wdyt?

fcakyon:
we could add a hook similar to on_server_start (that is, triggered on the webserver and not in the workers) that passes the number of active requests (incoming vs served). Wdyt?

Would be amazing!

aniketmaurya:
@vrdn-23 This middleware can track the active requests. @lantiga yeah we can have hooks `on_request` and `on_response` that is triggered in the endpoint functions. The will have access to this number of active requests through the headers as well. Is that what you meant?

```python
class RequestCountMiddleware(BaseHTTPMiddleware):
    """Adds a header to the response with the number of active requests."""

    def __init__(self, app: ASGIApp, active_counter: multiprocessing.Value) -> None:
        self.app = app
        self._active_requests = active_counter

    async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
        if scope["type"] != "http":
            await self.app(scope, receive, send)
            return
        with self._active_requests.get_lock():
            self._active_requests.value += 1
        await self.app(scope, receive, send)
        with self._active_requests.get_lock():
            self._active_requests.value -= 1
```

aniketmaurya:
@vrdn-23 @fcakyon you can track the active requests now! More details [here](https://lightning.ai/docs/litserve/features/full-control#track-active-requests).

fcakyon:
Fantastic news @aniketmaurya 🔥

PR #307, opened 2024-09-30
Dynamically grab port number

grantgasser:
<details>
  <summary><b>Before submitting</b></summary>

- [X] Was this discussed/agreed via a Github issue? (no need for typos and docs improvements) - **No, but discussed w/ Aniket on Slack**
- [X] Did you read the [contributor guideline](https://github.com/Lightning-AI/pytorch-lightning/blob/main/.github/CONTRIBUTING.md), Pull Request section?
- [ ] Did you make sure to update the docs?
- [ ] Did you write any new necessary tests?

</details>

⚠️ How does this PR impact the user? ⚠️
Now, running `python client.py` still works if user specifies port other than 8000 when calling `server.run()`. User doesn't need to know or have to change the port number generated in the `client.py` file.

## What does this PR do?

No issue created, but can do so for posterity's sake.

## PR review

Anyone in the community is free to review the PR once the tests have passed.
If we didn't discuss your PR in GitHub issues there's a high chance it will not be merged.

## Did you have fun?

Yes

PR #324, opened 2024-10-07
avoid client generation if already exist

aniketmaurya:
<details>

</details>

## What does this PR do?

avoid client generation if already exist. Caught by @bhimrazy https://github.com/Lightning-AI/LitServe/pull/307#issuecomment-2396490222

## PR review

## Did you have fun?

### DIFF
--- a/src/litserve/python_client.py
+++ b/src/litserve/python_client.py
@@ -1,3 +1,4 @@
+client_template = """
 # Copyright The Lightning AI team.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
@@ -13,5 +14,6 @@
 # limitations under the License.
 import requests
 
-response = requests.post("http://127.0.0.1:{{PORT}}/predict", json={"input": 4.0})
-print(f"Status: {response.status_code}\nResponse:\n {response.text}")
+response = requests.post("http://127.0.0.1:{PORT}/predict", json={{"input": 4.0}})
+print(f"Status: {{response.status_code}}\\nResponse:\\n {{response.text}}")
+"""
--- a/src/litserve/server.py
+++ b/src/litserve/server.py
@@ -27,7 +27,7 @@ from concurrent.futures import ThreadPoolExecutor
 from contextlib import asynccontextmanager
 from queue import Empty
 from typing import Callable, Dict, Optional, Sequence, Tuple, Union, List
-
+from litserve.python_client import client_template
 import uvicorn
 from fastapi import Depends, FastAPI, HTTPException, Request, Response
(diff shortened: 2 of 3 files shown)


lantiga (approved):

src/litserve/python_client.py
```diff
@@ -13,5 +14,6 @@
 # limitations under the License.
 import requests
 
-response = requests.post("http://127.0.0.1:{{PORT}}/predict", json={"input": 4.0})
-print(f"Status: {response.status_code}\nResponse:\n {response.text}")
+response = requests.post("http://127.0.0.1:{PORT}/predict", json={{"input": 4.0}})
+print(f"Status: {{response.status_code}}\\nResponse:\\n {{response.text}}")
+"""
```
Andrei-Aksionov, line 19: Just FIY, you can also use [Template](https://docs.python.org/3/library/string.html#template-strings). So you don't have to add double curly braces.
Here it's not needed, but if you will ever create a large template, it might come in handy.

It will be something like this:
```python
from string import Template

client_template = Template("""
# Copyright The Lightning AI team.
#
# Licensed under the Apache License, Version 2.0 (the "License");
# limitations under the License.
import requests

response = requests.post("http://127.0.0.1:${PORT}/predict", json={"input": 4.0})
print(f"Status: {response.status_code}\\nResponse:\\n {response.text}")
""")
```

and to use it
```python
client_template.substitute(PORT=port)
```

src/litserve/server.py
```diff
@@ -390,23 +390,19 @@ async def stream_predict(request: self.request_type) -> self.response_type:
...
         dest_path = os.path.join(os.getcwd(), "client.py")
 
-        try:
-            # Read the template code
-            with open(src_path) as f:
-                client_code = f.read()
-
-            # Replace the default port number with the user-specified port number
-            client_code = client_code.replace("{{PORT}}", str(port))
+        if os.path.exists(dest_path):
```
Andrei-Aksionov, line 395: But is it a correct behaviour?
If I changed the port I expect that the file should be regenerated with a new, correct one.

Or maybe I am missing something?

aniketmaurya: if you already have a `client.py` that you updated based on your server (let's say image classification). You won't want LitServe to replace the file. The generate client is supposed to be static template that people can quickly edit.

src/litserve/python_client.py
aniketmaurya: This is great! Didn't know about you. Thanks for letting me know.

aniketmaurya: I guess I can also use it for Dockerfile template

PR #342, opened 2024-10-23
Add custom healthcheck path

### LINKED ISSUE #341: Add custom path option for healthcheck
## 🚀 Feature

I'd like to be able to customize the path for the health check. Currently it defaults to `/health`, but I'd like to be able to add my own prefix to that, such as `/my_service/health` or any arbitrary value.

### Motivation

We use a shared load balancer with path-based routing, so we have to use app-specific prefixes for our paths, such as `/my_service_1/foo` and `/my_service_2/bar`. Because of this, the health check path also needs to start with that prefix.

### Pitch

Ideally I could add a `healthcheck_path` to the LitServer class initialization that would override the default path of `/health`.

### Alternatives

Currently I'm getting around this by using `server.app.add_api_route`, and adding a custom healthcheck method to my LitAPI class. In this method, I've tried using Requests to make a GET request to the local `/health` endpoint, but that actually causes LitServe to exit, so at the moment it just returns an HTTP 200 response, and doesn't actually check the readiness of the server.

akuma12:
<details>

- [ ] Did you make sure to update the docs? - I would if I knew where to do that.

</details>

## What does this PR do?

Fixes #341 by adding a `healthcheck_path` property to the LitServer class, defaulting to the current path of `/health`.

## PR review

## Did you have fun?
I did!

### DIFF
--- a/tests/test_lit_server.py
+++ b/tests/test_lit_server.py
@@ -14,6 +14,7 @@
 import asyncio
 import re
 import sys
+from time import sleep
 from unittest.mock import MagicMock, patch
 
 import pytest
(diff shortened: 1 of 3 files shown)


tests/test_lit_server.py
```diff
@@ -355,6 +356,19 @@ def test_custom_api_path():
...
+    with pytest.raises(ValueError, match="healthcheck_path must start with '/'. "):
+        LitServer(ls.test_examples.SimpleLitAPI(), healthcheck_path="customhealth")
+
+    server = LitServer(ls.test_examples.SimpleLitAPI(), healthcheck_path="/v1/custom_health")
+    url = server.healthcheck_path
+    with wrap_litserve_start(server) as server, TestClient(server.app) as client:
+        # Sleep a bit to ensure the server is ready
+        sleep(3)
+        response = client.get(url)
+        assert response.status_code == 200, "Server response should be 200 (OK)"
```
Borda, line 369: does all need to be in the `with` wrapper?

Borda (approved):

aniketmaurya (approved):
Thanks for your patience @akuma12! This looks great 🚀

PR #356, opened 2024-11-07
Fail fast when LitAPI.setup has error

### LINKED ISSUE #347: LitServe server doesn't crash when there is a failure in setup
### To Reproduce

Attach a [Lightning Studio](https://lightning.ai/studios) which is fully reproducible (code, dependencies, environment, etc...) to reproduce this:   

1. Create a [Studio](https://lightning.ai/studios).    
2. Reproduce the issue in the Studio.    
3. [Publish the Studio](https://lightning.ai/docs/overview/studios/publishing#how-to-publish).
4. Paste the Studio link here.    

```python

class DebugLitAPI(LitAPI):
    def setup(self, device):
        raise Exception

    def predict(self, inputs):
        return None

if __name__ == "__main__":
    api = DebugLitAPI()
    server = LitServer(api, accelerator='auto', devices='auto', workers_per_device=2)
    server.run(port=8000)
```

Instead, it should return exit code error 1.

### Environment
If you published a Studio with your bug report, we can automatically get this information. Otherwise, please describe:   

- PyTorch/Jax/Tensorflow Version (e.g., 1.0):
- OS (e.g., Linux):
- Python version:
- CUDA/cuDNN version:

aniketmaurya:
<details>

</details>

## What does this PR do?

Fixes #347 

LitServe doesn't crash when there is an error in the `LitAPI.setup` method. This PR makes LitServe fail when setup has error. 

## PR review

## Did you have fun?

### DIFF
--- a/src/litserve/loops.py
+++ b/src/litserve/loops.py
@@ -26,7 +26,7 @@ from starlette.formparsers import MultiPartParser
 from litserve import LitAPI
 from litserve.callbacks import CallbackRunner, EventTypes
 from litserve.specs.base import LitSpec
-from litserve.utils import LitAPIStatus, PickleableHTTPException
+from litserve.utils import LitAPIStatus, PickleableHTTPException, WorkerSetupStatus
 
 mp.allow_connection_pickling()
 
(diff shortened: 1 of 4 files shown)


src/litserve/loops.py
```diff
@@ -399,18 +399,23 @@ def inference_worker(
     max_batch_size: int,
     batch_timeout: float,
     stream: bool,
-    workers_setup_status: Dict[str, bool],
+    workers_setup_status: Dict[int, str],
```
Borda, line 402: can we have docs about the meaning or samples?

src/litserve/utils.py
```diff
@@ -78,3 +79,11 @@ async def call_after_stream(streamer: AsyncIterator, callback, *args, **kwargs):
         logger.exception(f"Error in streamer: {e}")
     finally:
         callback(*args, **kwargs)
+
+
+@dataclasses.dataclass
+class WorkerSetupStatus:
```
Borda, line 85: lets use strEnum https://github.com/Lightning-AI/utilities/blob/main/src/lightning_utilities/core/enums.py

src/litserve/loops.py
aniketmaurya: did you mean docstrings?

src/litserve/utils.py
aniketmaurya: what would be the benefit of using Enum over here?

ethanwharris (approved):

src/litserve/loops.py
Borda: Yes

src/litserve/utils.py
Borda: Becase you are using the data class as enum already

src/litserve/loops.py
aniketmaurya: great suggestion! although this is a private function so users won't be using it directly but might be useful for contributors.

PR #367, opened 2024-11-23
Feat support OpenAI embedding 

### LINKED ISSUE #305: Embedding model support with openai spec
Hi,
can i do the own custom embedding model deployment with litserve.? any document on this

bhimrazy:
<details>

</details>

## How does this PR impact users?
This PR addresses numerous user requests for an Embeddings API similar to [OpenAI’s](https://platform.openai.com/docs/api-reference/embeddings), making it easier for users to quickly get started with embeddings. By introducing this feature, users can now integrate embeddings in a plug-and-play manner, much like the Chat API, enhancing their overall experience and saving valuable setup time.

## What does this PR do?
Fixes #305.
- This PR adds support for an Embeddings specification similar to the [OpenAI Embeddings API](https://platform.openai.com/docs/api-reference/embeddings), allowing users to seamlessly work with embeddings in a way that’s simple and intuitive.

```python
# server.py
from typing import List

from litserve import LitServer, OpenAIEmbeddingSpec
from litserve.specs.openai_embedding import EmbeddingRequest
from fastembed import TextEmbedding


class EmbeddingAPI(ls.LitAPI):
    def setup(self, device):
        # Initialize the embedding model
        self.model = TextEmbedding(model_name="jinaai/jina-embeddings-v2-small-en")

    def decode_request(self, request: EmbeddingRequest, context: dict) -> List[str]:
        # Extract documents from the request
        documents = request.ensure_list()
       
       # context["prompt_tokens"] = ...
       # context["total_tokens"] = ...
        return documents

    def predict(self, documents) -> List[List[float]]:
        # Generate embeddings for the documents
        return list(self.model.embed(documents))

    def encode_response(self, output: List[List[float]], context: dict) -> dict:
        # Format the output as embeddings
        return {"embeddings": output, **context}


if __name__ == "__main__":
    api = EmbeddingAPI()
    server = ls.LitServer(api, spec=OpenAIEmbeddingSpec())
    # Run the server on port 8000
    server.run(port=8000)

```
Test API
```sh
curl http://localhost:8000/v1/embeddings \
  -H "Content-Type: application/json" \
  -d '{
    "input": "A beautiful sunset over the beach",
    "model": "jinaai/jina-embeddings-v2-small-en",
    "encoding_format": "float"
  }'
```
## PR review

## Did you have fun?

Absolutely! This was a fantastic learning experience and a joy to work on! 🚀🙌

### DIFF
--- a/src/litserve/specs/openai_embedding.py
+++ b/src/litserve/specs/openai_embedding.py
@@ -33,6 +33,10 @@ class EmbeddingRequest(BaseModel):
     dimensions: Optional[int] = None
     encoding_format: Literal["float"] = "float"
 
+    # TODO: Check if this might be a handy helper function
+    def get_input_as_list(self):
+        return self.input if isinstance(self.input, list) else [self.input]
+
 
 class Embedding(BaseModel):
     index: int
@@ -106,9 +110,7 @@ class OpenAIEmbeddingSpec(LitSpec):
         print("OpenAI Embedding Spec is ready.")
 
     def decode_request(self, request: EmbeddingRequest, context_kwargs: Optional[dict] = None) -> List[str]:
-        if isinstance(request.input, str):
-            return [request.input]
-        return request.input
+        return request.get_input_as_list()
 
     def encode_response(self, output: List[List[float]], context_kwargs: Optional[dict] = None) -> dict:
         return {
@@ -141,7 +143,7 @@ class OpenAIEmbeddingSpec(LitSpec):
         logger.debug(response)
 
         # TODO: Validate response and also confirm if UsageInfo should be default or not, as None is also a valid value
-        # maybe move this validate to setup
+        # maybe move this validate to setup for early validation
         self.validate_response(response)
 
         usage = UsageInfo(**response)
--- a/src/litserve/specs/openai_embedding.py
+++ b/src/litserve/specs/openai_embedding.py
@@ -130,7 +130,6 @@ class OpenAIEmbeddingSpec(LitSpec):
         }
 
     def validate_response(self, response: dict) -> None:
-        # Consider using the pattern: "What's wrong? Why it's wrong? How can I solve it?" when crafting error messages.
         if not isinstance(response, dict):
             raise ValueError(
                 "The response is not a dictionary."
--- a/src/litserve/specs/openai_embedding.py
+++ b/src/litserve/specs/openai_embedding.py
@@ -1,3 +1,16 @@
+# Copyright The Lightning AI team.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
 import asyncio
 import inspect
 import logging
--- a/src/litserve/specs/openai_embedding.py
+++ b/src/litserve/specs/openai_embedding.py
@@ -140,7 +140,8 @@ class OpenAIEmbeddingSpec(LitSpec):
 
         logger.debug(response)
 
-        # TODO: Validate response and also confirm if UsageInfo shoul be default or not, as None is also a valid value
+        # TODO: Validate response and also confirm if UsageInfo should be default or not, as None is also a valid value
+        # maybe move this validate to setup
         self.validate_response(response)
 
         usage = UsageInfo(**response)
--- a/src/litserve/specs/openai_embedding.py
+++ b/src/litserve/specs/openai_embedding.py
@@ -101,7 +101,8 @@ class OpenAIEmbeddingSpec(LitSpec):
         if inspect.isgeneratorfunction(lit_api.predict):
             raise ValueError(
                 "You are using yield in your predict method, which is used for streaming.",
-                "OpenAIEmbeddingSpec doesn't support streaming because producing embeddings is not a sequential operation.",
+                "OpenAIEmbeddingSpec doesn't support streaming because producing embeddings ",
+                "is not a sequential operation.",
                 "Please consider replacing yield with return in predict.\n",
                 EMBEDDING_API_EXAMPLE,
             )
@@ -110,7 +111,8 @@ class OpenAIEmbeddingSpec(LitSpec):
         if not is_encode_response_original and inspect.isgeneratorfunction(lit_api.encode_response):
             raise ValueError(
                 "You are using yield in your encode_response method, which is used for streaming.",
-                "OpenAIEmbeddingSpec doesn't support streaming because producing embeddings is not a sequential operation.",
+                "OpenAIEmbeddingSpec doesn't support streaming because producing embeddings ",
+                "is not a sequential operation.",
                 "Please consider replacing yield with return in encode_response.\n",
                 EMBEDDING_API_EXAMPLE,
             )
--- a/src/litserve/specs/openai_embedding.py
+++ b/src/litserve/specs/openai_embedding.py
@@ -102,6 +102,9 @@ class OpenAIEmbeddingSpec(LitSpec):
             raise response
 
         logger.debug(response)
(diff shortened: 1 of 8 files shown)


src/litserve/specs/openai_embedding.py
```diff
@@ -0,0 +1,99 @@
...
+
+
+class EmbeddingResponse(BaseModel):
+    data: List[Embedding]
+    model: str
+    object: Literal["list"] = "list"
+    usage: UsageInfo
+
+
+class OpenAIEmbeddingSpec(LitSpec):
```
aniketmaurya, line 55: probably we should inherit `OpenAISpec` and you can avoid some repetition in `setup` method.

src/litserve/specs/openai_embedding.py
```diff
@@ -0,0 +1,99 @@
...
+
+        super().setup(server)
+
+        lit_api = self._server.lit_api
+        if inspect.isgeneratorfunction(lit_api.predict):
+            raise ValueError("OpenAI embedding spec does not support streaming predictions.")
+
+        is_encode_response_original = lit_api.encode_response.__code__ is LitAPI.encode_response.__code__
+        if not is_encode_response_original and inspect.isgeneratorfunction(lit_api.encode_response):
+            raise ValueError("OpenAI embedding spec does not support streaming predictions.")
```
aniketmaurya, line 73: Let's put a sample example for users here in the error response:

```
class TestAPI(LitAPI):
    def setup(self, device):
        self.model = None

        return request

    def predict(self, x):
        return [np.random.rand(768).tolist()]

    def encode_response(self, response):
        return {"embeddings": response}
```

aniketmaurya (commented):
Looks good so far @bhimrazy 🚀 🚀 🚀

src/litserve/specs/openai_embedding.py
```diff
@@ -0,0 +1,155 @@
...
+
+
+class EmbeddingRequest(BaseModel):
+    input: Union[str, List[str]]
+    model: str
+    dimensions: Optional[int] = None
+    encoding_format: Literal["float"] = "float"
+
+    # TODO: Check if this might be a handy helper function
+    def get_input_as_list(self):
```
lantiga, line 37: We can name this `ensure_list`

src/litserve/specs/openai_embedding.py
lantiga: The code is short enough that I'd keep it flat.

Rather in the future we could have a `OpenAISpec` class and then specialized spec classes `OpenAIEmbeddingSpec` and `OpenAIChatCompletionSpec`.

We could then configure `OpenAISpec(api="chat-completion")` (default, this way current users will not be impacted) and `OpenAISpec(api="embedding")`, so the user would only ever see `OpenAISpec` and configure it with a string.

Not strictly needed right now and we can merge as is, but what's your take on this?

src/litserve/specs/openai_embedding.py
```diff
@@ -0,0 +1,155 @@
...
+        self.add_endpoint("/v1/embeddings", self.options_embeddings, ["GET"])
+
+    def setup(self, server: "LitServer"):  # noqa: F821
+        from litserve import LitAPI
+
+        super().setup(server)
+
+        lit_api = self._server.lit_api
+        if inspect.isgeneratorfunction(lit_api.predict):
+            raise ValueError(LITAPI_VALIDATION_MSG.format("predict is a generator"))
```
lantiga, line 104: The error message needs to be more friendly.
Consider using the pattern: "What's wrong? Why it's wrong? How can I solve it?" when crafting error messages.

In this case it could be something like "You are using `yield` in your `predict` method, which is used for streaming. OpenAIEmbeddingSpec doesn't support streaming because producing embeddings is not a sequential operation. Please consider replacing `yield` with `return` in `predict`."

src/litserve/specs/openai_embedding.py
```diff
@@ -0,0 +1,155 @@
...
+    def encode_response(self, output: List[List[float]], context_kwargs: Optional[dict] = None) -> dict:
+        return {
+            "embeddings": output,
+            "prompt_tokens": context_kwargs.get("prompt_tokens", 0),
+            "total_tokens": context_kwargs.get("total_tokens", 0),
+        }
+
+    def validate_response(self, response: dict) -> None:
+        if not isinstance(response, dict):
+            raise ValueError(RESPONSE_VALIDATION_MSG.format("response is not a dictionary"))
```
lantiga, line 124: See above re: error messages

src/litserve/specs/openai_embedding.py
```diff
@@ -0,0 +1,155 @@
...
+        response, status = self._server.response_buffer.pop(uid)
+
+        if status == LitAPIStatus.ERROR:
+            raise response
+
+        logger.debug(response)
+
+        # TODO: Validate response and also confirm if UsageInfo should be default or not, as None is also a valid value
+        # maybe move this validate to setup for early validation
+        self.validate_response(response)
```
lantiga, line 147: Since you are returning a dict, we don't have a lot of structure to work on, so moving to setup is not really an option. It's also a good opportunity to produce a good error message.

tests/e2e/test_e2e.py
```diff
@@ -326,3 +326,28 @@ def test_e2e_single_streaming():
     expected_values = [4.0, 8.0, 12.0]
     for i, output in enumerate(outputs):
         assert output["output"] == expected_values[i], f"Intermediate output {i} is not expected value"
+
+
+@e2e_from_file("tests/e2e/default_openai_embedding_spec.py")
```
lantiga, line 331: Another test with a little more coverage on the failure cases would help avoiding breakages in the future.

It's good to pad things in view of future changes, this is what tests are really for after all.

lantiga (commented):
Awesome contribution @bhimrazy, and a timely one too!

I added a few comments, we're very close

src/litserve/specs/openai_embedding.py
bhimrazy: Coincidentally, I was also thinking of something similar, breaking `ChatCompletionSpec` out from `OpenAISpec` as Luca has suggested. In the future, if we plan - it could be extended to include TTS, etc.

williamFalcon: yeah, let's not start nesting things like @lantiga suggested. We want to keep things as flat as possible and it is okay to have duplication (cc @aniketmaurya )

src/litserve/specs/openai_embedding.py
bhimrazy: Thank you for the clarification, @lantiga. This is very helpful! I'll make a note of this approach for crafting error messages and keep it as a reference for my other projects as well. Appreciate your guidance

src/litserve/specs/openai_embedding.py
```diff
@@ -56,36 +55,35 @@ class EmbeddingResponse(BaseModel):
...
+    def decode_request(self, request) -> List[str]:
         return request.input if isinstance(request.input, list) else [request.input]
 
-    def predict(self, x)-> List[List[float]]:
+    def predict(self, x) -> List[List[float]]:
         return np.random.rand(len(x), 768).tolist()
 
-    def encode_response(self, output)-> dict:
+    def encode_response(self, output) -> dict:
         return {"embeddings": output}
```
aniketmaurya: why do we need a list of strings to be returned from `decode_request`? Probably for simplicity of example, predict can do something like `np.random.rand(len([x] if isinstance(x, str) else x, 768).tolist()`

aniketmaurya: or use `request.ensure_list`

bhimrazy: Good point, @aniketmaurya. That approach definitely simplifies things and keeps the logic flexible.

bhimrazy: So , basically we can leave that up to user how they want to return from decode request: i.e either string to list of strings

aniketmaurya:
Looking great @bhimrazy! Another solid contribution 🚀 🙌 . 

One final thing - a test with batching would be amazing here.

bhimrazy:
> Looking great @bhimrazy! Another solid contribution 🚀 🙌 .
> 
> One final thing - a test with batching would be amazing here.

Thank you, @aniketmaurya ! 
I've added the batching test—hope it aligns with expectations.  😊

aniketmaurya (approved):
Great work @bhimrazy!

PR #387, opened 2024-12-06
add continuous batching loop 1/n

aniketmaurya:
## What does this PR do?

Allow users to quickly write a customizable continuous batching loop.

This is one of the first PRs to add a simple continuous batching loop and subjected to change till we are in `0.2.6.dev*`. I will be making multiple PRs to clean this up and add tests. 

### Example with vLLM engine

```python
from vllm import LLMEngine, EngineArgs, SamplingParams
from litserve.loops import ContinuousBatchingLoop, Output
from litserve.utils import LitAPIStatus
import litserve as ls

class LitVLLM(ls.LitAPI):
    def setup(self, device):
        engine_args = EngineArgs(model="meta-llama/Llama-3.2-1B", device=device, max_model_len=2048)
        self.engine = LLMEngine.from_engine_args(engine_args)

    def decode_request(self, request: dict) -> dict:
        return request

    def predict(self, x):
        # vLLM engine.generate is called in the loop
        yield x

    def encode_response(self, x: str):
        yield x


class vLLMLoop(ContinuousBatchingLoop):
    def __init__(self, max_sequence_length: int = 2048):
        super().__init__(max_sequence_length)
        self.uids = {}  # Maps vLLM request_id (str) to original uid

    def add_request(self, uid: str, request, lit_api: vLLMAPI, lit_spec) -> None:
        super().add_request(uid, request, lit_api, lit_spec)
        request_id = str(uid)
        sampling_params = SamplingParams(
            temperature=request.get("temperature", 0.0),
            max_tokens=request.get("max_tokens", 10)
        )
        lit_api.engine.add_request(
            request_id=request_id,
            prompt=request["prompt"],
            params=sampling_params
        )
        self.uids[request_id] = uid

    def step(self, prev_outputs, lit_api: vLLMAPI, lit_spec):
        request_outputs = lit_api.engine.step()
        outputs = []
        
        for request_output in request_outputs:
            if request_output.request_id not in self.uids:
                continue

            original_uid = self.uids[request_output.request_id]
            token_id = request_output.outputs[0].token_ids[-1]
            output = lit_api.engine.get_tokenizer().decode([token_id])
            outputs.append(Output(
                uid=original_uid,
                output=output,
                status=LitAPIStatus.OK
            ))

            if request_output.finished:
                outputs.append(Output(
                    uid=original_uid,
                    output="",
                    status=LitAPIStatus.FINISH_STREAMING
                ))
                del self.uids[request_output.request_id]

        return outputs

if __name__ == "__main__":
    loop = vLLMLoop()
    api = LitVLLM()
    server = ls.LitServer(api, loop=loop, max_batch_size=4, stream=True, timeout=-1)
    server.run()
```

Result of 4 concurrent requests: 

https://github.com/user-attachments/assets/08a8a88f-1179-4a71-9284-730282c2db0b

<details>

</details>

## PR review

## Did you have fun?

### DIFF
--- a/src/litserve/loops.py
+++ b/src/litserve/loops.py
@@ -618,61 +618,76 @@ class LitLoop(_BaseLoop):
 
 
 class ContinuousBatchingLoop(LitLoop):
-    def add_request(self, uid: str, request: Any, lit_api: LitAPI, lit_spec: Optional[LitSpec]) -> None:
-        raise NotImplementedError
+    def __init__(self, max_sequence_length: int = 2048):
+        super().__init__()
+        self.active_sequences: Dict[str, Dict] = {}  # uid -> {input, current_length, generated_tokens}
+        self.max_sequence_length = max_sequence_length
 
-    def step(self, lit_api: LitAPI, lit_spec: Optional[LitSpec]) -> List[Tuple[str, Tuple[Any, LitAPIStatus]]]:
-        raise NotImplementedError
+    def add_request(self, uid: str, request: Any, lit_api: LitAPI, lit_spec: Optional[LitSpec]) -> None:
+        """Add a new sequence to active sequences"""
+        self.active_sequences[uid] = {
+            "input": request,
+            "current_length": 0,
+            "generated_tokens": []
+        }
 
     def has_capacity(self, lit_api: LitAPI) -> bool:
-        raise NotImplementedError
+        """Check if we can add more sequences based on current batch"""
+        return len(self.active_sequences) < lit_api.max_batch_size
+
+    def step(self, lit_api: LitAPI, lit_spec: Optional[LitSpec]) -> List[Tuple[str, Tuple[Any, LitAPIStatus]]]:
+        """Process one token generation step for all active sequences"""
+        if not self.active_sequences:
+            return []
+
+        # Batch forward pass for all active sequences
+        inputs = [seq["input"] for seq in self.active_sequences.values()]
+        generated = [seq["generated_tokens"] for seq in self.active_sequences.values()]
+        
+        try:
+            # Assume lit_api.predict handles batched token generation
+            new_tokens = lit_api.predict(inputs, generated)
+            
+            responses = []
+            finished_uids = []
+
+            # Process each sequence's new token
+            for uid, token in zip(self.active_sequences.keys(), new_tokens):
+                seq = self.active_sequences[uid]
+                seq["generated_tokens"].append(token)
+                seq["current_length"] += 1
+
+                # Check completion conditions
+                is_finished = (
+                    token == lit_api.eos_token or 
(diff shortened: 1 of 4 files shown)


src/litserve/loops.py
```diff
@@ -37,10 +38,11 @@
     asyncio.set_event_loop_policy(uvloop.EventLoopPolicy())
 
 except ImportError:
-    print(
-        "uvloop is not installed. Falling back to the default asyncio event loop. "
-        "Please install uvloop for better performance using `pip install uvloop`."
-    )
+    if sys.platform != "win32":
```
ali-alshaar7, line 41: what if it is not on windows?

src/litserve/loops.py
```diff
@@ -573,6 +575,239 @@ def __call__(
...
+    status: LitAPIStatus
+
+
+class LitLoop(_BaseLoop):
+    def __init__(self):
+        self._context = {}
+
+    def get_batch_requests(self, lit_api: LitAPI, request_queue: Queue, max_batch_size: int, batch_timeout: float):
+        if max_batch_size <= 1:
+            raise ValueError("max_batch_size must be greater than 1")
```
ali-alshaar7, line 602: I think we should allow batch size 1 here, We may want to teak the batch size without opting for a different loop altogether

src/litserve/loops.py
```diff
@@ -573,6 +575,239 @@ def __call__(
...
+
+            # Process each sequence's new token
+            for uid, token in zip(self.active_sequences.keys(), new_tokens):
+                seq = self.active_sequences[uid]
+                seq["generated_tokens"].append(token)
+                seq["current_length"] += 1
+
+                # Check completion conditions
+                is_finished = token == lit_api.eos_token or seq["current_length"] >= self.max_sequence_length
+                is_finished = token == lit_api.eos_token or seq["current_length"] >= self.max_sequence_length
```
ali-alshaar7, line 683: doubled up why? Also, since we want to use this for both text and vision, should we be tied to `eos_token`? Id prefer if we let the end detection be configurable, since in LitGPT, Eos is not a single token, but a sequence of tokens

aniketmaurya: great catch. that was by mistake. you're right, we can let the user define how to detect `is_finished`.

src/litserve/loops.py
aniketmaurya: we have [get_request](https://github.com/Lightning-AI/LitServe/pull/387/files#diff-09cc07bb152f3c2d8b85404f472e5f4516835d1236155b431f48690d09ee790bR612) method to get the non-batched request. 

Two different methods because we need to check for timeout and notify users during batching.

ali-alshaar7 (approved):
LGTM!

Andrei-Aksionov (approved):

PR #392, opened 2024-12-10
Add image size control in openai.py

yaqiangsun:
When using litserve with a multimodal model （Qwen2-VL）, add resized_height and resized_width parameters to control the image size and prevent out-of-memory (OOM) issues. However, this feature does not work as expected.
:
```
messages=[
            {
                "role": "user",
                "content": [
                    {"type": "text", "text": prompt},
                    {
                        "type": "image",
                        "image_url": image_url,
                        "resized_height": 280,
                        "resized_width": 420
                    },
                ],
            }
        ],
```
This pr add the option params `resized_height` and `resized_width` to the `ImageContent`
```
class ImageContent(BaseModel):
    type: str
    image_url: Union[str, ImageContentURL]
    resized_height: Optional[int] = None
    resized_width: Optional[int] = None
```
And this modification does not affect the existing interface.
The key words  `resized_height` and `resized_width` are from Qwen2-VL package `qwen-vl-utils`:
[https://github.com/QwenLM/Qwen2-VL/blob/9806089cd7ebc341c3a77556e9cb18d2005853ad/qwen-vl-utils/src/qwen_vl_utils/vision_process.py#L105](https://github.com/QwenLM/Qwen2-VL/blob/9806089cd7ebc341c3a77556e9cb18d2005853ad/qwen-vl-utils/src/qwen_vl_utils/vision_process.py#L105)

## What does this PR do?

Fixes # (issue).

<details>

</details>

## PR review

## Did you have fun?

### DIFF
--- a/src/litserve/specs/openai.py
+++ b/src/litserve/specs/openai.py
@@ -75,6 +75,8 @@ class ImageContentURL(BaseModel):
 class ImageContent(BaseModel):
     type: str
     image_url: Union[str, ImageContentURL]
+    resized_height: Optional[int] = None
+    resized_width: Optional[int] = None
 
 
 class Function(BaseModel):


lantiga:
Thanks @yaqiangsun for the contribution. I agree with @aniketmaurya that we should stay as close as possible to the openai specification, so if it's missing there let's not add it here.

yaqiangsun:
@aniketmaurya @lantiga Thanks.
I completely understand your concern about maintaining compatibility with the OpenAI API. 
I'd like to clarify the PR. 

### The OpenAI API client have no influence with this PR 
This version maintains the core message while improving flow and readability.
OpenAI API has two part:
1. **Client** (e.g., the [Python package](https://github.com/openai/openai-python))
2. **OpenAI Cloud Serving System**

The second part, the OpenAI cloud serving system, does not support the parameters. However, adding these parameters will not require any changes to the client code itself. 
Here is the code sample:
```
stream = client.chat.completions.create(
        model=DEFAULT_MODEL,
        messages=[
            {
                "role": "user",
                "content": [
                    {"type": "text", "text": prompt},
                    {
                        "type": "image",
                        "image_url": image_url,
                        "resized_height": 280,
                        "resized_width": 420
                    },
                ],
            }
        ],
        stream=True,
        max_tokens=1024,
    )
```
Adding these parameters will not affect the client's functionality.

### This PR is fully compatible if keep strict with the OpenAI API 
Moreover, if the parameters `resized_height` and `resized_width` are added to the client, and the server doesn’t support them, the system will continue to function as it would without these parameters.
This pull request treats these parameters as **optional**, ensuring that if a user prefers to stay strictly within the OpenAI API's specifications, everything will still work smoothly.
Adding these parameters will not impact the server's behavior when adhering to the OpenAI API's strict guidelines.


# More LLM model support
The main purpose of `LitServe` is to help users build model serving systems, and OpenAI has already addressed the image size issue within its server system. I believe `LitServe` can remain compatible with the **OpenAI client** without needing to strictly follow the **OpenAI serving system**. This approach allows LitServe to focus on flexibility and model support while still integrating seamlessly with the OpenAI client.

yaqiangsun:
@aniketmaurya @lantiga Thanks.
I completely understand your concern about maintaining compatibility with the OpenAI API. 
I'd like to clarify the PR. 

### The OpenAI API client have no influence with this PR 
This version maintains the core message while improving flow and readability.
OpenAI API has two part:
1. **Client** (e.g., the [Python package](https://github.com/openai/openai-python))
2. **OpenAI Cloud Serving System**

The second part, the OpenAI cloud serving system, does not support the parameters. However, adding these parameters will not require any changes to the client code itself. 
Here is the code sample:
```
stream = client.chat.completions.create(
        model=DEFAULT_MODEL,
        messages=[
            {
                "role": "user",
                "content": [
                    {"type": "text", "text": prompt},
                    {
                        "type": "image",
                        "image_url": image_url,
                        "resized_height": 280,
                        "resized_width": 420
                    },
                ],
            }
        ],
        stream=True,
        max_tokens=1024,
    )
```
Adding these parameters will not affect the client's functionality.

### This PR is fully compatible if keep strict with the OpenAI API 
Moreover, if the parameters `resized_height` and `resized_width` are added to the client, and the server doesn’t support them, the system will continue to function as it would without these parameters.
This pull request treats these parameters as **optional**, ensuring that if a user prefers to stay strictly within the OpenAI API's specifications, everything will still work smoothly.
Adding these parameters will not impact the server's behavior when adhering to the OpenAI API's strict guidelines.


# More LLM model support
The main purpose of `LitServe` is to help users build model serving systems, and OpenAI has already addressed the image size issue within its server system. I believe `LitServe` can remain compatible with the **OpenAI client** without needing to strictly follow the **OpenAI serving system**. This approach allows LitServe to focus on flexibility and model support while still integrating seamlessly with the OpenAI client.

aniketmaurya:
Hi @yaqiangsun, while adding this field doesn’t affect OpenAI client compatibility, it introduces a feature that is "nice to have" but might be handled differently by various users. For example, clients could resize images themselves instead of relying on the server or enforce a global image size. Adding new fields to the payload for different use cases might make it bloated very quickly.

However, you can still include this field for your specific use case without modifying the LitServe source code. To do this, you can override the OpenAISpec.chat_completion method and define your custom [Pydantic request schema](https://lightning.ai/docs/litserve/features/request-response-format#pydantic-input).

PR #403, opened 2025-01-08
integrate zmq

aniketmaurya:
## What does this PR do?

<details>

</details>

Faster process communication with `zmq`.

**TODO:** A follow up PR to add a proxy and support multiple inference worker processes and multiple uvicorn processes. 

## PR review

## Did you have fun?

### DIFF
--- a/src/litserve/loops/base.py
+++ b/src/litserve/loops/base.py
@@ -20,6 +20,7 @@ from abc import ABC
 from queue import Empty, Queue
 from typing import Any, Dict, List, Optional, Tuple, Union
 
+import zmq
 from starlette.formparsers import MultiPartParser
 
 from litserve import LitAPI
@@ -128,6 +129,8 @@ class _BaseLoop(ABC):
 
     """
 
+    socket: zmq.Socket | None = None
+
     def pre_setup(self, lit_api: LitAPI, spec: Optional[LitSpec]):
         pass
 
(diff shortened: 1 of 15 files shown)


ethanwharris (approved):

src/litserve/loops/base.py
```diff
@@ -20,6 +20,7 @@
 from queue import Empty, Queue
 from typing import Any, Dict, List, Optional, Tuple, Union
 
+import zmq
 from starlette.formparsers import MultiPartParser
```
aniketmaurya, line 25: blocking merge, Luca has some thoughts. 
```suggestion

```

lantiga (commented):
The PR looks good in general, but not abstracting away the interprocess communication mechanism makes the code more complex (you have a socket rather than a queue so that has more opaque semantics), also a lot of details are guarded by `if .. else` conditions

I would take this opportunity to create a base class that takes care of the communication, and two concrete classes. You can then instantiate them in LitServer based on the one you want and you don't have to deal with conditionals and keep the code easily consumable.

Clear semantics is particularly important for people implementing loops, we need to keep it simple.

aniketmaurya:
agree with your points @lantiga, a better place to put these would be inside [this put_response](https://github.com/Lightning-AI/LitServe/blob/501639b099b5ebcdabc3ada329faabff91a52561/src/litserve/loops/base.py#L244) method which can be used everywhere else. I will also create an encapsulation to hide these socket details for the next PR which would enable zmq for multiple workers.

williamFalcon (approved):

aniketmaurya:
creating followup as per above suggestion.

PR #407, opened 2025-01-09
Fix: Add Callback Events and Align Hooks in Streaming Loop

bhimrazy:
## What does this PR do?

- [x] Adds missing **before** and **after** callback events for encoding responses in the streaming loop.  
- [x] Also aligns the hooks (`after_predict` > `after_encode_response`) and ensures they run only after the generator completes.  
  - This ensures that the `after` hooks are executed only once the generator finishes processing all tasks.

## PR review

## Did you have fun?


src/litserve/loops/streaming_loops.py
```diff
@@ -70,11 +70,14 @@ def run_streaming_loop(
             )
             callback_runner.trigger_event(EventTypes.AFTER_PREDICT, lit_api=lit_api)
 
+            callback_runner.trigger_event(EventTypes.BEFORE_ENCODE_RESPONSE, lit_api=lit_api)
```
aniketmaurya, line 73: this one is a bit tricky since `y_enc_gen` is a generator and the values are actually not materialized till it goes through [these](https://github.com/Lightning-AI/LitServe/pull/407/files#diff-8b336fb2d3f02a375742713860907ac68109570f4dfe1366c34e35aa6ef97168R81-R84) lines. so can't add it here.

aniketmaurya: maybe we can show a warning to users when this callback hook is used with streaming.

bhimrazy: Sure, @aniketmaurya. 
It seems that even the predict hook does not execute as expected since `predict` is also a generator. Additionally, it appears that all of the last four events are executed before the `predict` function runs.
  <details>
  <summary><b>Script I used to test</b></summary>
  
  ```python
  from litserve import Callback, LitAPI, LitServer
  
  
  class SimpleStreamAPI(LitAPI):
          self.model = lambda x: x
  
          return request["input"]
  
      def predict(self, x):
          print(f"Predicting {x}")
          for i in range(4):
              yield self.model(i)
  
          for out in output:
              yield {"output": out}
  
  
  class LogCallback(Callback):
      def on_before_predict(self, *args, **kwargs):
          print("Before prediction")
  
      def on_after_predict(self, *args, **kwargs):
          print("After prediction")
  
      def on_before_encode_response(self, *args, **kwargs):
          print("Before encoding response")
  
      def on_after_encode_response(self, *args, **kwargs):
          print("After encoding response")
  
  
  if __name__ == "__main__":
      api = SimpleStreamAPI()
      server = LitServer(api, stream=True, callbacks=[LogCallback()])
      server.run(port=8000)
  ```
  </details>

One potential approach could be to move the "after event" part to execute after the generator has completed. However, this might not align semantically(in code) with the idea of executing it immediately after the respective functions.

or Alternatively, as you suggested, we could simply display a warning to users when this callback hook is used with streaming.

Looking forward to your thoughts on this  😊.

aniketmaurya: thanks for looking into it @bhimrazy! we can align all the hooks (after_predict>after_encode_response) and run them after the generator stops and add this case in the documentation.

bhimrazy: Sure

aniketmaurya (approved):
Looks great @bhimrazy! let's goo 🚀

aniketmaurya (approved):

PR #427, opened 2025-02-13
fix: reponse format JSONSchema key fix

Lucaz0619:
## What does this PR do?

This PR fixes openai.BadRequestError 400 - missing required parameter related to response_format.json_schema.schema. The issue was caused by a naming mismatch in JSONSchema parsing when handling response formats. This is a simple fix that ensures proper parameter inclusion.

<details>

- [x ] Was this discussed/agreed via a Github issue? (no need for typos and docs improvements)
- [v] Did you read the [contributor guideline](https://github.com/Lightning-AI/pytorch-lightning/blob/main/.github/CONTRIBUTING.md), Pull Request section?
- [v] Did you make sure to update the docs?

</details>

## How does this PR impact the user?
✅ Resolves incorrect request failures due to mismatch JSONSchema parameters. 

## Did you have fun?
Absolutely!

### DIFF
--- a/src/litserve/specs/openai.py
+++ b/src/litserve/specs/openai.py
@@ -117,7 +117,7 @@ class ResponseFormatJSONObject(BaseModel):
 class JSONSchema(BaseModel):
     name: str
     description: Optional[str] = None
-    schema: Optional[Dict[str, object]] = Field(None, alias="schema")
+    schema_: Optional[Dict[str, object]] = Field(None, alias="schema")
     strict: Optional[bool] = False
 
 
--- a/src/litserve/specs/openai.py
+++ b/src/litserve/specs/openai.py
@@ -117,7 +117,7 @@ class ResponseFormatJSONObject(BaseModel):
 class JSONSchema(BaseModel):
     name: str
     description: Optional[str] = None
-    schema_def: Optional[Dict[str, object]] = Field(None, alias="schema")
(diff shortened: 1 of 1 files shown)


src/litserve/specs/openai.py
```diff
@@ -117,7 +117,7 @@ class ResponseFormatJSONObject(BaseModel):
 class JSONSchema(BaseModel):
     name: str
     description: Optional[str] = None
-    schema_def: Optional[Dict[str, object]] = Field(None, alias="schema")
+    schema: Optional[Dict[str, object]] = Field(None, alias="schema")
```
bhimrazy, line 120: I think this will have an issue with overwriting the parent attribute of BaseModel.

Borda:
Does it happen for and version or just the latest release?

src/litserve/specs/openai.py
aniketmaurya: nice catch @bhimrazy! yeah we need to use the alias in this case. [here](https://github.com/openai/openai-python/blob/3f8d8205ae41c389541e125336b0ae0c5e437661/src/openai/types/shared/response_format_json_schema.py#L27) is the reference to the official OpenAI input format.

aniketmaurya: @Lucaz0619 I added a commit to your PR to apply the above change.

aniketmaurya (approved):

PR #430, opened 2025-02-18
Feat: add custom health check logic

### LINKED ISSUE #426: Allow users to define a custom health check logic
----

## 🚀 Feature

Right now the health check endpoint returns "ok" when all the processes has started. There could be scenario when the LitAPI depends on another API/service (such as check if Ollama has pulled the model in background) and health-check need to account for the liveliness of that service too. 

It would be great to allow users to provide a custom health logic:

```py
class MyAPI(ls.LitAPI):
    def health(self):
        if SOMETHING:
            return A
        else: 
            return B
```

bhimrazy:
## What does this PR do?
This PR introduces the ability to define custom health check logic in LitAPI, allowing users to tailor health checks to their specific needs.

Fixes #426.

<details>

</details>

## PR review

## Did you have fun?

### DIFF
--- a/src/litserve/api.py
+++ b/src/litserve/api.py
@@ -143,3 +143,7 @@ class LitAPI(ABC):
 
     def has_capacity(self) -> bool:
         raise NotImplementedError("has_capacity is not implemented")
+
+    def health(self) -> bool:
+        # TODO: Discuss what this should return
+        return True
--- a/src/litserve/api.py
+++ b/src/litserve/api.py
@@ -145,13 +145,13 @@ class LitAPI(ABC):
         raise NotImplementedError("has_capacity is not implemented")
 
     def health(self) -> bool:
-        """
-        Check the additional health status of the API.
+        """Check the additional health status of the API.
 
         This method is used in the /health endpoint of the server to determine the health status.
         Users can extend this method to include additional health checks specific to their application.
 
         Returns:
             bool: True if the API is healthy, False otherwise.
+
         """
         return True
--- a/src/litserve/api.py
+++ b/src/litserve/api.py
@@ -145,5 +145,13 @@ class LitAPI(ABC):
         raise NotImplementedError("has_capacity is not implemented")
 
     def health(self) -> bool:
-        # TODO: Discuss what this should return
+        """
+        Check the additional health status of the API.
+
+        This method is used in the /health endpoint of the server to determine the health status.
+        Users can extend this method to include additional health checks specific to their application.
+
+        Returns:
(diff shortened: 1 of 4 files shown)


src/litserve/api.py
```diff
@@ -143,3 +143,7 @@ def has_active_requests(self) -> bool:
 
     def has_capacity(self) -> bool:
         raise NotImplementedError("has_capacity is not implemented")
+
+    def health(self) -> bool:
+        # TODO: Discuss what this should return
```
justusschock, line 148: This needs proper documentation :)

tests/test_simple.py
```diff
@@ -103,6 +115,38 @@ def test_workers_health_custom_path(use_zmq):
...
+        response = client.get("/health")
+        assert response.status_code == 503
+        assert response.text == "not ready"
+
+        time.sleep(1)
+        response = client.get("/health")
+        assert response.status_code == 503
+        assert response.text == "not ready"
+
+        time.sleep(3)
```
justusschock, line 139: let's only sleep 1 here and sleep 4 below. Depending on the time the test and client takes for setup, this might become flaky otherwise.

justusschock (commented):
Looking great. Few comments :)

tests/test_simple.py
bhimrazy: Sure

src/litserve/api.py
bhimrazy: you mean the docstring, right?  

btw, @aniketmaurya, I also wanted to check if we keep it as `NotImplemented` or a default value with `True` is fine?

aniketmaurya: i think default `True` would be fine for this.

tests/test_simple.py
```diff
@@ -53,6 +53,20 @@ def setup(self, device):
...
+        self.model = lambda x: x**2
+        time.sleep(2)
+
+    def health(self) -> bool:
+        # Custom health check logic
+        if not hasattr(self, "_start_time"):
+            self._start_time = time.time()
+
+        # Healthy if the server has been running for at least 5 seconds
+        return time.time() - self._start_time >= 5
```
aniketmaurya, line 67:
```suggestion
        # Health check passes after 5 seconds from the first time it is called. 
        if not hasattr(self, "_start_time"):
            self._start_time = time.time()
        return time.time() - self._start_time >= 5
```

`setup` is called in a subprocess and `health` is called in the main process

aniketmaurya (approved):
Looks great @bhimrazy!

PR #435, opened 2025-02-23
fix: update type hints for _default_unbatch and _spec attributes in LitAPI class 

bhimrazy:
### **What does this PR do?**  
**fix:** Update type hints for `_default_unbatch` and `_spec` attributes in `LitAPI` class  

**Changes:**  
- `callable` (lowercase) is not a valid type hint; replaced with `Callable` from `typing`.  
-  Update type hints for `_default_unbatch` and `_spec` attributes
- Added a validation check to ensure `_default_unbatch` is set before calling it.  

## PR review

## Did you have fun?

### DIFF
--- a/src/litserve/api.py
+++ b/src/litserve/api.py
@@ -15,7 +15,7 @@ import json
 import warnings
 from abc import ABC, abstractmethod
 from queue import Queue
-from typing import Optional
+from typing import Callable, Optional
 
 from pydantic import BaseModel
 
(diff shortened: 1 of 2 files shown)


src/litserve/api.py
```diff
@@ -76,6 +76,8 @@ def _unbatch_stream(self, output_stream):
 
     def unbatch(self, output):
         """Convert a batched output to a list of outputs."""
+        if self._default_unbatch is None:
+            raise ValueError("Unbatch function is not set.")
```
aniketmaurya, line 80: Let's make the error more user friendly, something like - "Default implementation for `LitAPI.unbatch` method was not found, please implement LitAPI.unbatch method."

aniketmaurya (commented):
Looking great! Added a comment.

src/litserve/api.py
bhimrazy: Sure done. Also added a test case for the `ValueError`.

ethanwharris (approved):

PR #439, opened 2025-02-27
Feature: 🚀 Add Audio Content Support to OpenAISpec Request

### LINKED ISSUE #438: Add `input_audio` Support to OpenAISpec Request
### Summary
Add support for `input_audio` in OpenAISpec to align with multimodal that accept audio inputs. This will extend the spec to handle audio data, making it more compatible with OpenIAI API.
Reference: [OpenAI API Reference](https://platform.openai.com/docs/api-reference/chat/create?lang=python)

Image

### Motivation
As models like [Phi-4-multimodal-instruct](https://huggingface.co/microsoft/Phi-4-multimodal-instruct) support audio, adding this to OpenAISpec will keep it up-to-date and versatile, allowing seamless audio input handling.

bhimrazy:
## What does this PR do?
**🚀 Add Audio Content Support to OpenAISpec Request**

This PR introduces support for the Audio Content in the OpenAISpec request. 

This will extend the spec to handle input audio data, making it more compatible with OpenIAI API.
Reference: [OpenAI API Reference](https://platform.openai.com/docs/api-reference/chat/create?lang=python)

Image

Fixes #438

## PR review

## Did you have fun?
Yes, I did 😊.

### DIFF
--- a/src/litserve/specs/openai.py
+++ b/src/litserve/specs/openai.py
@@ -78,8 +78,7 @@ class ImageContent(BaseModel):
 
 
 class InputAudio(BaseModel):
-    data: str
-    """Base64 encoded audio data."""
+    data: str  # base64 encoded audio data.
 
     format: Literal["wav", "mp3"]
 
--- a/src/litserve/specs/openai.py
+++ b/src/litserve/specs/openai.py
@@ -79,13 +79,11 @@ class ImageContent(BaseModel):
 
 class InputAudio(BaseModel):
     data: str  # base64 encoded audio data.
-
     format: Literal["wav", "mp3"]
 
 
 class AudioContent(BaseModel):
     type: Literal["input_audio"]
-
     input_audio: InputAudio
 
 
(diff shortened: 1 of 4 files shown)


src/litserve/specs/openai.py
```diff
@@ -77,6 +77,19 @@ class ImageContent(BaseModel):
     image_url: Union[str, ImageContentURL]
 
 
+class InputAudio(BaseModel):
+    data: str
+    """Base64 encoded audio data."""
```
justusschock, line 82: this is the way to document the entire class (in which case it'd have to be above data). If you want to document the data attribute only, the docstring should be a inline comment

justusschock (approved):

src/litserve/specs/openai.py
bhimrazy: Sure @justusschock, thanks so much for the clarification! 
I’ve updated it accordingly.

aniketmaurya (approved):

PR #478, opened 2025-05-07
async support for streaming loop

aniketmaurya:
## What does this PR do?

Follow up to https://github.com/Lightning-AI/LitServe/pull/477, users would be able to run streaming APIs with async. 

```python
import litserve as ls

class AsyncAPI(ls.LitAPI):
        self.model = lambda x: x

    async def decode_request(self, request):
        return request["input"]

    async def predict(self, x):
        for i in range(10):
            yield self.model(i)

    async def encode_response(self, output):
        for out in output:
            yield {"output": out}

if __name__ == "__main__":
    api = AsyncAPI(enable_async=True)
    server = ls.LitServer(api, stream=True)
    server.run(port=8000)
```

<details>

</details>

## PR review

## Did you have fun?

### DIFF
--- a/src/litserve/loops/base.py
+++ b/src/litserve/loops/base.py
@@ -47,9 +47,16 @@ def _inject_context(context: Union[List[dict], dict], func, *args, **kwargs):
 
 async def _async_inject_context(context: Union[List[dict], dict], func, *args, **kwargs):
     sig = inspect.signature(func)
+    is_async_gen = inspect.isasyncgenfunction(func)
+
     if "context" in sig.parameters:
-        return await func(*args, **kwargs, context=context)
-    return await func(*args, **kwargs)
+        result = (
+            await func(*args, **kwargs, context=context) if not is_async_gen else func(*args, **kwargs, context=context)
+        )
+    else:
+        result = await func(*args, **kwargs) if not is_async_gen else func(*args, **kwargs)
+
+    return result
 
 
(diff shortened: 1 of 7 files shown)


k223kim (approved):

src/litserve/loops/base.py
```diff
@@ -277,39 +284,55 @@ def pre_setup(self, lit_api: LitAPI, spec: Optional[LitSpec]):
 
         original = lit_api.unbatch.__code__ is LitAPI.unbatch.__code__
         if not lit_api.stream and any([
-            inspect.isgeneratorfunction(lit_api.predict),
-            inspect.isgeneratorfunction(lit_api.encode_response),
+            inspect.isgeneratorfunction(lit_api.predict) or inspect.isasyncgenfunction(lit_api.predict),
+            inspect.isgeneratorfunction(lit_api.encode_response) or inspect.isasyncgenfunction(lit_api.encode_response),
```
justusschock, line 288: nit, no need to `or` them here. the global `any` is basically a `or` already

justusschock (approved):

src/litserve/loops/base.py
aniketmaurya: thanks! nice catch

aniketmaurya: leaving it this way though for readability

PR #479, opened 2025-05-08
Enable true concurrency in async loop

bhimrazy:
## What does this PR do?

Follow-up to #477. 
The async loop now processes tasks concurrently, so new requests don't have to wait for previous ones to finish.

## PR review

## Did you have fun?

### DIFF
--- a/tests/test_lit_server.py
+++ b/tests/test_lit_server.py
@@ -605,11 +605,13 @@ async def test_concurrent_async_inference():
                 end_time = monotonic()
                 output = resp.json()["output"]
                 input_duration = output**0.5
-                completions.append({
-                    "input": input_duration,
-                    "output": output,
-                    "elapsed": end_time - start_times[input_duration],
-                })
+                completions.append(
+                    {
+                        "input": input_duration,
+                        "output": output,
+                        "elapsed": end_time - start_times[input_duration],
+                    }
+                )
 
             # The order of completion should be shortest sleep first
             completed_order = [item["input"] for item in completions]
@@ -621,6 +623,6 @@ async def test_concurrent_async_inference():
             # TODO: discuss to update this test, as time based seems to be flaky in CI
             # max_elapsed = max(item["elapsed"] for item in completions)
             # assert max_elapsed < sum(sleep_durations), (
-            #     f"Total elapsed time ({max_elapsed:.2f}s) should be less than sum of sleeps ({sum(sleep_durations)}s), "
-            #     "indicating concurrent execution."
+            #     "Total elapsed time ({max_elapsed:.2f}s) should be less than sum of sleeps"
+            #     f" ({sum(sleep_durations)}s), indicating concurrent execution."
             # )
--- a/tests/test_lit_server.py
+++ b/tests/test_lit_server.py
@@ -601,4 +601,4 @@ async def test_concurrent_async_inference():
                 assert resp.json()["output"] == 25.0
 
             # All requests should finish in just over 4s, plus some overhead
-            assert elapsed < 4 + 2, f"Expected all requests to finish in just over 4s, but took {elapsed:.2f}s."
+            assert elapsed < 4 + 4, f"Expected all requests to finish in just over 4s, but took {elapsed:.2f}s."
--- a/tests/test_lit_server.py
+++ b/tests/test_lit_server.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 import asyncio
 import sys
-from time import sleep
+from time import monotonic, sleep
 from unittest.mock import MagicMock, patch
 
 import pytest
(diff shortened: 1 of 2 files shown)


tests/test_lit_server.py
```diff
@@ -562,3 +562,65 @@ async def test_async_litapi():
...
+class TestSleepAsyncLitAPI(ls.LitAPI):
+    def setup(self, device):
+        self.model = None
+
+    async def decode_request(self, request):
+        return request["input"]
+
+    async def predict(self, x):
+        # sleep x time to simulate a long-running task
+        await asyncio.sleep(x)
```
aniketmaurya, line 576: We can have a fixed sleep time here and send concurrent (maybe 10) requests from the client. 

To assert that the server is concurrent and not sequential, we should get the response for all the requests within 4 seconds instead of `num_requests x 4`. 

```suggestion
        await asyncio.sleep(4)
```

tests/test_lit_server.py
```diff
@@ -562,3 +562,65 @@ async def test_async_litapi():
...
+                f"Expected completion order [1.0, 2.0, 4.0], got {completed_order}"
+            )
+
+            # The total elapsed time should be just over the longest sleep, not the sum
+            # TODO: discuss to update this test, as time based seems to be flaky in CI
+            # max_elapsed = max(item["elapsed"] for item in completions)
+            # assert max_elapsed < sum(sleep_durations), (
+            #     "Total elapsed time ({max_elapsed:.2f}s) should be less than sum of sleeps"
+            #     f" ({sum(sleep_durations)}s), indicating concurrent execution."
+            # )
```
aniketmaurya, line 626: let's simplify the test as per above suggestion

aniketmaurya (changes requested):
Looking great @bhimrazy! let's simplify the test a bit and we are good to go. 

Also, would be nice follow up PR to ensure that errors from the background task (`asyncio.create_task`) are propagated to users properly.

tests/test_lit_server.py
```diff
@@ -13,7 +13,7 @@
 # limitations under the License.
 import asyncio
 import sys
-from time import sleep
+from time import sleep, time
```
aniketmaurya, line 16:
```suggestion
from time import sleep
```

tests/test_lit_server.py
```diff
@@ -562,3 +562,42 @@ async def test_async_litapi():
...
+    """Test that async API processes requests concurrently (not sequentially)."""
+    api = TestSleepAsyncLitAPI(enable_async=True)
+    server = LitServer(api)
+    with wrap_litserve_start(server) as server:
+        async with LifespanManager(server.app) as manager, AsyncClient(
+            transport=ASGITransport(app=manager.app), base_url="http://test"
+        ) as ac:
+            num_requests = 10
+            tasks = [ac.post("/predict", json={"input": 5.0}, timeout=10) for _ in range(num_requests)]
+            start = time()
```
aniketmaurya, line 594:
```suggestion
            start = time.perf_counter()
```

tests/test_lit_server.py
```diff
@@ -562,3 +562,42 @@ async def test_async_litapi():
...
+            responses = await asyncio.gather(*tasks)
+            elapsed = time() - start
```
aniketmaurya, line 596:
```suggestion
            elapsed = time.perf_counter() - start
```

tests/test_lit_server.py
```diff
@@ -562,3 +562,42 @@ async def test_async_litapi():
...
+
+            for resp in responses:
+                assert resp.status_code == 200
+                assert resp.json()["output"] == 25.0
+
+            # All requests should finish in just over 4s, plus some overhead
+            assert elapsed < 4 + 4, f"Expected all requests to finish in just over 4s, but took {elapsed:.2f}s."
```
aniketmaurya, line 603:
```suggestion
            assert elapsed < 4 + 2, f"Expected all requests to finish in just over 4s, but took {elapsed:.2f}s."
```

aniketmaurya (approved):

tests/test_lit_server.py
```diff
@@ -562,3 +563,42 @@ async def test_async_litapi():
...
+            start = time.perf_counter()
+            responses = await asyncio.gather(*tasks)
+            elapsed = time.perf_counter() - start
+
+            for resp in responses:
+                assert resp.status_code == 200
+                assert resp.json()["output"] == 25.0
+
+            # All requests should finish in just over 4s, plus some overhead
+            assert elapsed < 4 + 2, f"Expected all requests to finish in just over 4s, but took {elapsed:.2f}s."
```
aniketmaurya, line 604: ideally 2 seconds of extra buffer should be fine but CI is taking 7.8 seconds on average to process 10 requests which is still faster than (10x4) 40 seconds
```suggestion
            assert elapsed < 4 + 4, f"Expected all requests to finish in just over 4s, but took {elapsed:.2f}s."
```

PR #482, opened 2025-05-09
Save references for async tasks to prevent garbage collection during execution.

bhimrazy:
## What does this PR do ?

Follow-up to #479
Store references to asynchronous tasks to prevent them from being garbage-collected during execution. This change enhances the reliability of concurrent request processing.

> Ref: https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task
> asyncio.create_task is a massive footgun because it can silently make tasks disappear: https://x.com/mitsuhiko/status/1920384040005173320

### DIFF
--- a/src/litserve/loops/simple_loops.py
+++ b/src/litserve/loops/simple_loops.py
@@ -197,6 +197,7 @@ class SingleLoop(DefaultLoop):
         callback_runner: CallbackRunner,
     ):
         async def process_requests():
+            pending_tasks = set()
             while True:
                 try:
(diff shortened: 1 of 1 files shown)


aniketmaurya (approved):
Thanks @bhimrazy looks great! Can you also add a test for `_process_single_request`, maybe in a separate PR if you want as we are low on coverage here.

bhimrazy:
> Thanks @bhimrazy looks great! Can you also add a test for `_process_single_request`, maybe in a separate PR if you want as we are low on coverage here.

Sure @aniketmaurya

PR #485, opened 2025-05-11
Add tests for async loop processing

bhimrazy:
## What does this PR do?
Follow-up to #482

Introduce tests for `_process_single_request` and `_run_single_loop_with_async` for processing async requests.
> increases the test coverage overall from 85% to 86%.

image

### DIFF
--- a/tests/test_loops.py
+++ b/tests/test_loops.py
@@ -158,10 +158,10 @@ def test_run_single_loop_with_async(async_loop_args, monkeypatch):
     # Patch kill to do nothing in test
     monkeypatch.setattr(loop, "kill", lambda: None)
 
-    try:  # noqa: SIM105
+    # Expected to break the loop in test
+    import contextlib
+    with contextlib.suppress(KeyboardInterrupt):
         loop._run_single_loop_with_async(lit_api_mock, None, requests_queue, mock_transport, NOOP_CB_RUNNER)
-    except KeyboardInterrupt:
-        pass  # Expected to break the loop in test
 
     response = asyncio.get_event_loop().run_until_complete(mock_transport.areceive(consumer_id=0))
     assert response == ("uuid-123", ({"output": 1}, ls.utils.LitAPIStatus.OK))
--- a/tests/test_loops.py
+++ b/tests/test_loops.py
@@ -160,6 +160,7 @@ def test_run_single_loop_with_async(async_loop_args, monkeypatch):
 
     # Expected to break the loop in test
     import contextlib
+
     with contextlib.suppress(KeyboardInterrupt):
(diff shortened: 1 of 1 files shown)


tests/test_loops.py
```diff
@@ -88,6 +130,45 @@ def test_single_loop(loop_args):
...
+
+    loop = SingleLoop()
+
+    # Patch kill to do nothing in test
+    monkeypatch.setattr(loop, "kill", lambda: None)
+
+    try:  # noqa: SIM105
+        loop._run_single_loop_with_async(lit_api_mock, None, requests_queue, mock_transport, NOOP_CB_RUNNER)
+    except KeyboardInterrupt:
+        pass  # Expected to break the loop in test
```
aniketmaurya, line 164:
```suggestion
    # Expected to break the loop in test
    import contextlib
    with contextlib.suppress(KeyboardInterrupt):
        loop._run_single_loop_with_async(lit_api_mock, None, requests_queue, mock_transport, NOOP_CB_RUNNER)
```

tests/test_loops.py
```diff
@@ -70,6 +71,47 @@ def loop_args():
...
+
+    def get(self, timeout=None):
+        if self._sentinel_seen:
+            raise Empty  # Simulate queue being empty after sentinel
+        item = super().get(timeout=timeout)
+        # Sentinel: (None, None, None, None)
+        if item == (None, None, None, None):
+            self._sentinel_seen = True
+            raise KeyboardInterrupt  # Triggers loop exit in your code
+        return item
```
aniketmaurya, line 87: love this! in future, maybe we can add something directly in the Loops for test mode so that we don't have to do this everywhere.

tests/test_loops.py
```diff
@@ -88,6 +130,45 @@ def test_single_loop(loop_args):
...
+
+
+def test_run_single_loop_with_async(async_loop_args, monkeypatch):
+    mock_transport = MockMPQueueTransport(num_consumers=2)
+    lit_api_mock, requests_queue = async_loop_args
+
+    loop = SingleLoop()
+
+    # Patch kill to do nothing in test
+    monkeypatch.setattr(loop, "kill", lambda: None)
```
aniketmaurya, line 159: nice, I think we might have to look at kill later on. It has introduced some weird behavior while terminating the server. maybe only allow it for windows.

aniketmaurya (approved):

tests/test_loops.py
bhimrazy: sure.

tests/test_loops.py
bhimrazy: sure.

PR #488, opened 2025-05-13
Enable true concurrency in async streaming loop and add tests

bhimrazy:
## What does this PR do ?

Follow-up to #478
The async streaming loop now processes tasks concurrently, so new requests don't have to wait for previous ones to finish.

Also, adds tests for async streaming functionality and concurrent request handling.

### DIFF
--- a/src/litserve/loops/streaming_loops.py
+++ b/src/litserve/loops/streaming_loops.py
@@ -185,6 +185,7 @@ class StreamingLoop(DefaultLoop):
         callback_runner: CallbackRunner,
     ):
         async def process_requests():
+            pending_tasks = set()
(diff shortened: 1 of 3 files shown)


src/litserve/loops/streaming_loops.py
```diff
@@ -185,9 +185,13 @@ def run_streaming_loop_async(
         callback_runner: CallbackRunner,
     ):
         async def process_requests():
+            event_loop = asyncio.get_running_loop()
+            pending_tasks = set()
```
aniketmaurya, line 189:
```suggestion
            pending_tasks = set()

```

PR #499, opened 2025-05-16
fix: OpenAI Spec validations to work with async LitAPI

bhimrazy:
## What does this PR do ?

Introduce validation messages for async LitAPI functions in the OpenAISpec and add tests to ensure compliance with async requirements.

### DIFF
--- a/src/litserve/specs/openai.py
+++ b/src/litserve/specs/openai.py
@@ -351,6 +351,15 @@ class OpenAISpec(LitSpec):
                 raise ValueError(ASYNC_LITAPI_VALIDATION_MSG.format("encode_response is not a generator"))
 
         else:
+            for method in ["decode_request", "predict", "encode_response"]:
+                method_obj = getattr(lit_api, method)
+                if asyncio.iscoroutinefunction(method_obj) or inspect.isasyncgenfunction(method_obj):
+                    raise ValueError(
+                        f"'{method}' is defined as async/coroutine, but 'enable_async' is not set in LitAPI. "
+                        "Please set 'enable_async=True' in your LitAPI class or "
+                        "use synchronous (non-async) methods instead."
+                    )
+
             if not inspect.isgeneratorfunction(lit_api.predict):
                 raise ValueError(LITAPI_VALIDATION_MSG.format("predict is not a generator"))
 
--- a/src/litserve/specs/openai.py
+++ b/src/litserve/specs/openai.py
@@ -344,6 +344,9 @@ class OpenAISpec(LitSpec):
         is_encode_response_original = lit_api.encode_response.__code__ is LitAPI.encode_response.__code__
 
         if lit_api.enable_async:
+            if not asyncio.iscoroutinefunction(lit_api.decode_request):
+                raise ValueError(ASYNC_LITAPI_VALIDATION_MSG.format("decode_request is not a coroutine"))
+
             if not inspect.isasyncgenfunction(lit_api.predict):
                 raise ValueError(ASYNC_LITAPI_VALIDATION_MSG.format("predict is not a generator"))
 
--- a/src/litserve/specs/openai.py
+++ b/src/litserve/specs/openai.py
@@ -332,20 +332,21 @@ class OpenAISpec(LitSpec):
     def pre_setup(self, lit_api: "LitAPI"):
         from litserve import LitAPI
 
-        lit_api_is_async = lit_api.enable_async
-        # validations for async
-        if lit_api_is_async and not (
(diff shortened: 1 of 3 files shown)


src/litserve/specs/openai.py
```diff
@@ -280,12 +341,21 @@ def stream(self):
...
         from litserve import LitAPI
 
-        if not inspect.isgeneratorfunction(lit_api.predict):
-            raise ValueError(LITAPI_VALIDATION_MSG.format("predict is not a generator"))
-
         is_encode_response_original = lit_api.encode_response.__code__ is LitAPI.encode_response.__code__
-        if not is_encode_response_original and not inspect.isgeneratorfunction(lit_api.encode_response):
-            raise ValueError(LITAPI_VALIDATION_MSG.format("encode_response is not a generator"))
+
+        if lit_api.enable_async:
```
aniketmaurya, line 346:
```suggestion
        if lit_api.enable_async:

            if not inspect.isasyncgenfunction(lit_api.decode_request):
                raise ValueError(ASYNC_LITAPI_VALIDATION_MSG.format("decode_request is not a generator"))

```

aniketmaurya (approved):
Looks great @bhimrazy! Let's also add a validation case when the methods are async but `enable_async` was not set.

tests/test_specs.py
```diff
@@ -457,3 +457,66 @@ async def test_openai_embedding_spec_with_batching(openai_embedding_request_data
...
+    with wrap_litserve_start(server) as server:
+        async with LifespanManager(server.app) as manager, AsyncClient(
+            transport=ASGITransport(app=manager.app), base_url="http://test"
+        ) as ac:
+            resp = await ac.post("/v1/chat/completions", json=openai_request_data, timeout=10)
+            assert resp.status_code == 200, "Status code should be 200"
+
+            assert resp.json()["choices"][0]["message"]["content"] == "This is a sample response", (
+                "LitAPI predict response should match with the generated output"
+            )
```
Borda, line 522:
```suggestion
     assert resp.status_code == 200, "Status code should be 200"

     assert resp.json()["choices"][0]["message"]["content"] == "This is a sample response", (
                "LitAPI predict response should match with the generated output"
            )
```

tests/test_specs.py
```diff
@@ -457,3 +457,66 @@ async def test_openai_embedding_spec_with_batching(openai_embedding_request_data
...
+        return request
+
+    async def predict(self, x):
+        for token in self.sentence:
+            yield token
+
+    async def encode_response(self, output_stream, context):
+        for output in output_stream:
+            yield {"role": "assistant", "content": output}
+
```
aniketmaurya, line 508: let's also test scenario when decode is not implemented by the user

aniketmaurya (changes requested):

tests/test_specs.py
bhimrazy: sure. done

tests/test_specs.py
```diff
@@ -510,6 +510,23 @@ def test_openai_asyncapi_enable_async_flag_validation():
...
+        yield "This is a generated output"
+
+    async def encode_response(self, output):
+        yield {"role": "assistant", "content": output}
+
+
+@pytest.mark.asyncio
+def test_openai_asyncapi_decode_not_implemented():
+    with pytest.raises(ValueError, match=r"LitAPI\(enable_async=True\) requires all methods to be coroutines\."):
+        ls.LitServer(DecodeNotImplementedAsyncOpenAILitAPI(enable_async=True), spec=OpenAISpec())
```
aniketmaurya, line 527: @bhimrazy i think this should pass, LitSpec should has a default async implementation.

bhimrazy: > @bhimrazy I think this should pass — `LitSpec` should provide a default async implementation.

@aniketmaurya, do you have any suggestions on the implementation?

Currently, it seems even if we add a separate default async version in `LitSpec` (with a slightly different name), we still need to explicitly use it in `LitAPI`. I'm not sure if we can have both async and non-async methods with the same name.

https://github.com/Lightning-AI/LitServe/blob/a8787a365e9d6e53e84d157414b8a20ad42e50a6/src/litserve/api.py#L87-L91

aniketmaurya: maybe we can "asyncify" the method if `enable_async` is set and methods are not implemented as coroutines.

aniketmaurya: with a warning of course.

bhimrazy: Oh, sure.
I just created a follow-up PR to asyncify `decode_request`: https://github.com/Lightning-AI/LitServe/pull/502
Should we do the same for `encode_request` as well?

PR #500, opened 2025-05-18
fix OpenAI embedding spec for batching

aniketmaurya:
## What does this PR do?

This PR addresses two key issues related to request decoding and response formatting when using the OpenAIEmbeddingSpec:

**1. Request Input Wrapping:**
The decode_request function always wraps request.input in a list, which breaks the downstream logic for certain input formats. This PR ensures inputs are handled more intelligently based on their structure.

**2. Output Shape Handling:**
When predict returns a 1D tensor or NumPy array (e.g., shape (128,)), the OpenAIEmbeddingSpec attempts to wrap each element individually into the Embedding class, leading to incorrect behavior. This PR corrects that by ensuring the output is treated as a single embedding vector when appropriate.

✅ Example Use Case
The following example previously failed due to the issues above but now works as expected after this fix:

```python
from sentence_transformers import SentenceTransformer
import litserve as ls

class EmbeddingsAPI(ls.LitAPI):
    def setup(self, device):
        self.model = SentenceTransformer("all-MiniLM-L6-v2", device=device)

    def predict(self, inputs):
        # Fix for nested list inputs
        if isinstance(inputs[0], list) and len(inputs[0]) == 1:
            inputs = inputs[0]
        return self.model.encode(inputs)

if __name__ == "__main__":
    api = EmbeddingsAPI(max_batch_size=2, batch_timeout=2)
    server = ls.LitServer(api, spec=ls.OpenAIEmbeddingSpec())
    server.run(port=8000)
```

After this PR, users will be able to run embedding models with LitServe and the OpenAIEmbeddingSpec seamlessly, even when the model outputs a 1D vector or the input structure is slightly irregular.

<details>

</details>

## PR review

## Did you have fun?

### DIFF
--- a/src/litserve/specs/base.py
+++ b/src/litserve/specs/base.py
@@ -25,6 +25,7 @@ class LitSpec:
         self._endpoints = []
 
         self._server: LitServer = None
+        self._max_batch_size = 1
 
     @property
     def stream(self):
--- a/src/litserve/specs/openai_embedding.py
+++ b/src/litserve/specs/openai_embedding.py
@@ -78,14 +78,15 @@ Please follow the example below for guidance on how to use the OpenAI Embedding 
 ```python
 import numpy as np
 from typing import List
+from litserve.specs import EmbeddingRequest
 from litserve import LitAPI, OpenAIEmbeddingSpec
 
 class TestAPI(LitAPI):
     def setup(self, device):
         self.model = None
 
-    def decode_request(self, request) -> List[str]:
-        return request.ensure_list()
+    def decode_request(self, request: EmbeddingRequest) -> List[str]:
+        return request.input
 
     def predict(self, x) -> List[List[float]]:
         return np.random.rand(len(x), 768).tolist()
@@ -136,7 +137,7 @@ class OpenAIEmbeddingSpec(LitSpec):
         print("OpenAI Embedding Spec is ready.")
 
     def decode_request(self, request: EmbeddingRequest, context_kwargs: Optional[dict] = None) -> List[str]:
-        return request.ensure_list()
+        return request.input
 
     def encode_response(self, output: List[List[float]], context_kwargs: Optional[dict] = None) -> dict:
         usage = {
--- a/src/litserve/specs/openai_embedding.py
+++ b/src/litserve/specs/openai_embedding.py
@@ -78,25 +78,44 @@ Please follow the example below for guidance on how to use the OpenAI Embedding 
 ```python
 import numpy as np
 from typing import List
-from litserve.specs import EmbeddingRequest
-from litserve import LitAPI, OpenAIEmbeddingSpec
+from litserve.specs import OpenAIEmbeddingSpec, EmbeddingRequest
+import litserve as ls
 
-class TestAPI(LitAPI):
+class TestAPI(ls.LitAPI):
     def setup(self, device):
         self.model = None
 
-    def decode_request(self, request: EmbeddingRequest) -> List[str]:
-        return request.input
+    def predict(self, inputs) -> List[List[float]]:
+        # inputs is a string
+        return np.random.rand(1, 768).tolist()
 
-    def predict(self, x) -> List[List[float]]:
-        return np.random.rand(len(x), 768).tolist()
-
-    def encode_response(self, output) -> dict:
-        return {"embeddings": output}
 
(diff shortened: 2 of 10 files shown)


src/litserve/specs/base.py
```diff
@@ -25,6 +25,7 @@ def __init__(self):
         self._endpoints = []
 
         self._server: LitServer = None
+        self._max_batch_size = 1
```
KaelanDt, line 28: is there a reason to hard code the batch size to 1?

tests/test_openai_embedding.py
```diff
@@ -0,0 +1,183 @@
...
+    with wrap_litserve_start(server) as server:
+        async with LifespanManager(server.app) as manager, AsyncClient(
+            transport=ASGITransport(app=manager.app), base_url="http://test"
+        ) as ac:
+            resp = await ac.post("/v1/embeddings", json=openai_embedding_request_data, timeout=10)
+            assert resp.status_code == 200, "Status code should be 200"
+            assert resp.json()["object"] == "list", "Object should be list"
+            assert resp.json()["data"][0]["index"] == 0, "Index should be 0"
+            assert len(resp.json()["data"]) == 1, "Length of data should be 1"
+            assert len(resp.json()["data"][0]["embedding"]) == 768, "Embedding length should be 768"
```
KaelanDt, line 87: is there not a way to get the values here without hardcoding?

tests/test_specs.py
```diff
@@ -337,128 +324,6 @@ async def test_fail_http(openai_request_data):
             assert res.text == '{"detail":"test LitAPI.predict error"}'
 
 
-@pytest.mark.asyncio
```
KaelanDt, line 340: what is the rationale for moving this test to another file?

KaelanDt (approved):
small nits, overall looks good to me!

src/litserve/specs/openai_embedding.py
```diff
@@ -152,8 +188,60 @@ def _validate_response(self, response: dict) -> None:
...
+    def _handle_embedding_response(
+        self, embeddings: Union[List, "np.ndarray", "torch.Tensor", "List[List[float]]"], num_items: int = 1
+    ) -> List[Embedding]:
+        ndim = None
+        if "torch" in sys.modules:
+            import torch
+
+            if isinstance(embeddings, torch.Tensor):
+                ndim = embeddings.ndim
+        if "numpy" in sys.modules:
```
Borda, line 200:
```suggestion
        elif "numpy" in sys.modules:
```

src/litserve/specs/openai_embedding.py
```diff
@@ -152,8 +188,60 @@ def _validate_response(self, response: dict) -> None:
...
+            import numpy as np
+
+            if isinstance(embeddings, np.ndarray):
+                ndim = embeddings.ndim
+
+        # expand_dims for torch.Tensor or np.ndarray
+        if ndim == 1:
+            embeddings = embeddings[None, :]
+
+        if ndim is not None:
```
Borda, line 210:
```suggestion
        elif ndim is not None:
```

src/litserve/specs/base.py
aniketmaurya: it's default to 1

tests/test_openai_embedding.py
aniketmaurya: these are hardcoded since it's a test. We want to make sure it's 768 since the API implements 768 embedding dim and similarly for other items it should be static in every condition.

tests/test_specs.py
aniketmaurya: moved OpenAIEmbedding based tests to test_openai_embedding.py

src/litserve/specs/openai_embedding.py
aniketmaurya: we want to apply this independent of the above condition

src/litserve/specs/openai_embedding.py
aniketmaurya: we want to check both. so it wont be 'else'

tests/test_openai_embedding.py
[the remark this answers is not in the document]
aniketmaurya: right. can move it later

PR #502, opened 2025-05-19
chore: asyncify default decode_request method if not overridden by user

bhimrazy:
## What does this PR do?

This PR is a follow-up to #499.
It ensures that the default `decode_request` method is converted to an asynchronous function if the user has not overridden it.
This allows consistent async behavior across implementations and improves compatibility with async-based request handling.

## PR review

## Did you have fun?

### DIFF
--- a/src/litserve/api.py
+++ b/src/litserve/api.py
@@ -54,6 +54,14 @@ class LitAPI(ABC):
 
         # Asyncify default decode_request if not implemented by user when enable_async is True
         if self.enable_async and not asyncio.iscoroutinefunction(self.decode_request):
+            warnings.warn(
+                "'decode_request' is not implemented as an async coroutine. "
+                "Since 'enable_async=True', the default implementation will be wrapped to run asynchronously. "
+                "To avoid this warning and ensure proper async support, "
+                "please override 'decode_request' as 'async def decode_request(...)'.",
+                UserWarning,
+                stacklevel=2,
+            )
             self.decode_request = self._asyncify(self.decode_request)
 
         self._validate_async_methods()
--- a/src/litserve/api.py
+++ b/src/litserve/api.py
@@ -16,6 +16,7 @@ import inspect
 import json
 import warnings
 from abc import ABC, abstractmethod
+from functools import wraps
 from queue import Queue
 from typing import Callable, Optional
 
@@ -50,7 +51,21 @@ class LitAPI(ABC):
         self.max_batch_size = max_batch_size
         self.batch_timeout = batch_timeout
         self.enable_async = enable_async
+
+        # asyncify default decode_request if not implemented by user when enable_async is True
+        if self.enable_async and not asyncio.iscoroutinefunction(self.decode_request):
+            self.decode_request = self._asyncify(self.decode_request)
+
         self._validate_async_methods()
+
+    def _asyncify(self, func):
+        """Wrap a function to be async if enable_async is True."""
+
+        @wraps(func)
+        async def async_wrapper(*args, **kwargs):
+            return await asyncio.to_thread(func, *args, **kwargs)
+
+        return async_wrapper
 
     def _validate_async_methods(self):
         """Validate that async methods are properly implemented when enable_async is True."""
--- a/src/litserve/api.py
+++ b/src/litserve/api.py
@@ -52,14 +52,14 @@ class LitAPI(ABC):
         self.batch_timeout = batch_timeout
         self.enable_async = enable_async
 
-        # asyncify default decode_request if not implemented by user when enable_async is True
+        # Asyncify default decode_request if not implemented by user when enable_async is True
         if self.enable_async and not asyncio.iscoroutinefunction(self.decode_request):
             self.decode_request = self._asyncify(self.decode_request)
 
         self._validate_async_methods()
 
     def _asyncify(self, func):
-        """Wrap a function to be async if enable_async is True."""
+        """Wrap a synchronous function to be async using asyncio.to_thread."""
 
         @wraps(func)
         async def async_wrapper(*args, **kwargs):
--- a/tests/test_specs.py
+++ b/tests/test_specs.py
@@ -522,8 +522,8 @@ class DecodeNotImplementedAsyncOpenAILitAPI(ls.LitAPI):
 
 
 @pytest.mark.asyncio
-def test_openai_asyncapi_decode_not_implemented():
-    with pytest.raises(ValueError, match=r"LitAPI\(enable_async=True\) requires all methods to be coroutines\."):
+def test_openai_asyncapi_decode_not_implemented_warning():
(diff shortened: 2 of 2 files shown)


src/litserve/api.py
```diff
@@ -50,8 +51,30 @@ def __init__(self, max_batch_size: int = 1, batch_timeout: float = 0.0, enable_a
...
+        if self.enable_async and not asyncio.iscoroutinefunction(self.decode_request):
+            warnings.warn(
+                "'decode_request' is not implemented as an async coroutine. "
+                "Since 'enable_async=True', the default implementation will be wrapped to run asynchronously. "
+                "To avoid this warning and ensure proper async support, "
+                "please override 'decode_request' as 'async def decode_request(...)'.",
+                UserWarning,
+                stacklevel=2,
+            )
+            self.decode_request = self._asyncify(self.decode_request)
```
aniketmaurya, line 65: needs to be applied to the encode_response as well which would be an async generator. We will have to run the generator in a thread and yield the output as soon as they are generated in that case. 

For predict, we can enforce users to define as async rather than asyncify.

aniketmaurya (commented):
Good start @bhimrazy! Added few comments - 

- Let's create a decorator which can be applied to the OpenAI Spec methods to keep the code clean.

src/litserve/api.py
aniketmaurya: feel free to take a look at [this branch](https://github.com/Lightning-AI/LitServe/compare/asyncify?expand=1)

bhimrazy: Sure and thank you @aniketmaurya

src/litserve/api.py
```diff
@@ -96,34 +125,35 @@ def __init__(
...
-        async def decode_request(self, request, **kwargs):
-            return request
-        async def predict(self, x, **kwargs):
-            return x
-        async def encode_response(self, output, **kwargs):
-            return output
+            self._asyncify_methods()
+            self._validate_async_methods()
+
+    def _asyncify_methods(self):
```
Borda, line 133:
```suggestion
    def _asyncify_methods(self, methods: tuple=("decode_request", "encode_response")):
```

src/litserve/api.py
```diff
@@ -96,34 +125,35 @@ def __init__(
...
+                setattr(self, method, asyncify(method_obj))
 
-Streaming example:
-    class MyStreamingAPI(LitAPI):
-        async def predict(self, x, **kwargs):
-            for i in range(10):
-                await asyncio.sleep(0.1)  # simulate async work
-                yield f"Token {i}: {x}"
-""")
+    def _validate_async_methods(self):
```
Borda, line 146:
```suggestion
    def _validate_async_methods(self, methods: tuple = ("decode_request", "predict", "encode_response")):
```

src/litserve/utils.py
```diff
@@ -200,3 +203,71 @@ def set_trace_if_debug(debug_env_var="LITSERVE_DEBUG", debug_env_var_value="1"):
...
+    while True:
+        item = await queue.get()
+        if item is None:
+            break
+        elif isinstance(item, Exception):
+            raise item
+        yield item
+
+
+def asyncify(func):
```
Borda, line 238:
```suggestion
def asyncify(func: Callable):
```

src/litserve/api.py
```diff
@@ -96,34 +125,35 @@ def __init__(
...
+            self._validate_async_methods()
+
+    def _asyncify_methods(self, methods: tuple = ("decode_request", "encode_response")):
+        """Asyncify methods if enable_async is True and not already async."""
+        for method in methods:
+            method_obj = getattr(self, method)
+            if not (asyncio.iscoroutinefunction(method_obj) or inspect.isasyncgenfunction(method_obj)):
+                warnings.warn(
+                    f"enable_async set to True but {method} is not a coroutine or async generator. "
+                    "LitServe will asyncify this method."
```
aniketmaurya, line 140:
```suggestion
                    "LitServe will automatically convert this method to an 'async function' a.k.a 'coroutine'"
```

src/litserve/specs/openai.py
```diff
@@ -345,14 +347,23 @@ def pre_setup(self, lit_api: "LitAPI"):
         is_encode_response_original = lit_api.encode_response.__code__ is LitAPI.encode_response.__code__
 
         if lit_api.enable_async:
+            # warning for decode_request and encode_response
```
aniketmaurya, line 350:
```suggestion
```

src/litserve/specs/openai.py
```diff
@@ -345,14 +347,23 @@ def pre_setup(self, lit_api: "LitAPI"):
...
             if not asyncio.iscoroutinefunction(lit_api.decode_request):
-                raise ValueError(ASYNC_LITAPI_VALIDATION_MSG.format("decode_request is not a coroutine"))
+                logger.info("decode_request is not a coroutine function. LitServe will asyncify it.")
```
aniketmaurya, line 352: LitAPI should have already asyncfied it so let's keep this error
```suggestion
                raise ValueError(ASYNC_LITAPI_VALIDATION_MSG.format("decode_request is not a coroutine"))
```

src/litserve/specs/openai.py
```diff
@@ -345,14 +347,23 @@ def pre_setup(self, lit_api: "LitAPI"):
...
+            if not inspect.isasyncgenfunction(lit_api.encode_response):
+                logger.info("encode_response is not an async generator. LitServe will asyncify it.")
```
aniketmaurya, line 354: same as above

tests/test_specs.py
aniketmaurya, line 1: let's also add a success test which sends request and receives a response with asyncify.

bhimrazy:
Looks like the `asyncify` decorator is breaking function identity, causing `PicklingError`  — likely due to the function losing its reference to the original class/method. We might need to preserve the original `__qualname__` or explore using `functools.update_wrapper` or `wrapt` to retain picklability. Open to suggestions on the best approach here!


```python
    def dump(obj, file, protocol=None):
        '''Replacement for pickle.dump() using ForkingPickler.'''
>       ForkingPickler(file, protocol).dump(obj)
E       _pickle.PicklingError: Can't pickle <function TestMixedAsyncLitAPI.decode_request at 0x154c774c0>: it's not the same object as tests.test_lit_server.TestMixedAsyncLitAPI.decode_request

../../../.local/share/uv/python/cpython-3.12.5-macos-aarch64-none/lib/python3.12/multiprocessing/reduction.py:60: PicklingError
```

aniketmaurya:
thanks for this awesome effort @bhimrazy but it looks like this issue was solved as a side-effect of [this PR](https://github.com/Lightning-AI/LitServe/pull/552). I realized it while doing [this PR](https://github.com/Lightning-AI/LitServe/pull/553).

bhimrazy:
Closing this as the issue got resolved — thanks for the reviews and learnings! 🙏

Issue #492, opened 2025-05-13
Multiple endpoints roadmap


aniketmaurya:
## 🚀 Support for Multiple Endpoints in LitServe

We are introducing support for multiple endpoints in **LitServe**, which will require minor architectural adjustments to better separate concerns between the core components.

### ✨ Proposed Usage

Here’s an example of what the new API could look like:

```python
api1 = VisionEncoderAPI(batch_size=4, api_path="/ap1")
api2 = TextEncoderAPI(batch_size=8, api_path="/path2)

def resize(...):
   ...

server = LitServer([
  api1,
  api2,
  {"/random": resize}
])
server.run()
```

### 🔍 Design Goals

To cleanly support this functionality, we need to ensure that responsibilities are clearly divided between the two main entities:

#### `LitAPI`
- Manages properties that pertain to the logic of individual APIs, such as:
  - `batch_size`
  - `loop`
  - `stream`
  - `api_path`

#### `LitServer`
- Handles server-level configuration, such as:
  - Endpoint routing

#### Shared Responsibility (`LitAPI` + `LitServer`)
- Callback handling
- Streaming functionality

### 📌 Action Items

In [PR #468](https://github.com/Lightning-AI/LitServe/pull/468), we successfully moved the `batch_size` parameter from `LitServer` to `LitAPI` to align with this separation of concerns. We need to apply a similar refactoring for other attributes:

- [x] Move `loop` from `LitServer` to `LitAPI` - https://github.com/Lightning-AI/LitServe/pull/512
- [x] Move `stream` argument from LitServer to LitAPI - https://github.com/Lightning-AI/LitServe/pull/512
- [x] Move `api_path` - https://github.com/Lightning-AI/LitServe/pull/512
- [x] Remove `LitServer` argument from `LitSpec.setup`. Connect the queues and response buffer with a connector instead. 
- [x] Add multiple LitAPI support - https://github.com/Lightning-AI/LitServe/pull/513
- [x] Allow streaming and non-streaming together. - https://github.com/Lightning-AI/LitServe/pull/518
- [ ] Scale multiple APIs separately as described [here](https://github.com/Lightning-AI/LitServe/issues/514).
- [ ] Multiple endpoints with a single LitAPI (shared setup)

ywh-my:
Hello, dear developer. May I ask which branch contains the code for this feature?

aniketmaurya:
[...]

PR #513, opened 2025-05-25
Support multiple LitAPIs for inference process and endpoints

aniketmaurya:
## What does this PR do?

Users will be able to run multiple APIs in a single server. 

Part of https://github.com/Lightning-AI/LitServe/issues/492

### Example
```python
import litserve as ls

class SentimentAnalysisAPI(ls.LitAPI):
    def setup(self, device):
        self.model = pipeline("sentiment-analysis", model="stevhliu/my_awesome_model", device=device)

    def decode_request(self, request: dict):
        return request["text"]

    def predict(self, text):
        return self.model(text)[0]

class TextGenerationAPI(ls.LitAPI):
    def setup(self, device):
        self.generator = pipeline("text-generation", model="gpt2", device=device)

    def decode_request(self, request: dict):
        return request["prompt"]

    def predict(self, prompt):
        return self.generator(prompt)[0]["generated_text"]

if __name__ == "__main__":
    sentiment_api = SentimentAnalysisAPI(api_path="/classify-text")
    chat_api = TextGenerationAPI(api_path="/v1/chat/completions")

    server = ls.LitServer([sentiment_api, chat_api])
    server.run(port=8000)
```

image

<details>

</details>

## PR review

## Did you have fun?

### DIFF
--- a/src/litserve/server.py
+++ b/src/litserve/server.py
@@ -162,8 +162,24 @@ class _LitAPIConnector:
             if not self.lit_apis:  # Check if the iterable is empty
                 raise ValueError("lit_apis must not be an empty iterable")
             self._detect_path_collision()
+            self._check_mixed_streaming_configuration()
         else:
             raise ValueError(f"lit_apis must be a LitAPI or an iterable of LitAPI, but got {type(lit_apis)}")
+
+    def _check_mixed_streaming_configuration(self):
+        """Ensure consistent streaming configuration across all endpoints.
+
+        Streaming must be either enabled for all endpoints or disabled for all. Mixing streaming and non-streaming
+        endpoints is currently not supported.
+
+        """
+        streams_enabled = [api.stream for api in self.lit_apis]
+        if any(streams_enabled) and not all(streams_enabled):
+            raise ValueError(
+                "Inconsistent streaming configuration: all endpoints must either "
+                "enable streaming or disable it. "
+                "Mixed configurations are not supported yet."
+            )
 
     def _detect_path_collision(self):
         paths = {"/health": "LitServe healthcheck", "/info": "LitServe info"}
--- a/tests/test_multiple_endpoints.py
+++ b/tests/test_multiple_endpoints.py
@@ -57,3 +57,14 @@ def test_reserved_paths():
     api2 = InferencePipeline(name="api2", api_path="/info")
     with pytest.raises(ValueError, match="api_path /health is already in use by LitServe healthcheck"):
         ls.LitServer([api1, api2])
+
+
+def test_check_mixed_streaming_configuration():
+    api1 = InferencePipeline(name="api1", api_path="/api1", stream=True)
+    api2 = InferencePipeline(name="api2", api_path="/api2", stream=False)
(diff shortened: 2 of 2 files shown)


src/litserve/server.py
```diff
@@ -352,16 +365,16 @@ def launch_inference_worker(self, num_uvicorn_servers: int):
...
-                    self.lit_api,
-                    self.lit_api._spec,
+                    lit_api,
+                    lit_api._spec,
                     device,
                     worker_id,
                     self.request_queue,
                     self._transport,
-                    self.lit_api.stream,
+                    lit_api.stream,
```
bhimrazy, line 374: In the follow-ups, I think we can also look into reducing some of the parameters here.

By the way, (just for learning) — what would be considered better practice in this case: keeping things flat as we have now, or directly accessing attributes from `lit_api` within the inference worker?

src/litserve/server.py
```diff
@@ -624,7 +637,12 @@ def run(
         elif api_server_worker_type is None:
             api_server_worker_type = "process"
 
-        manager, inference_workers = self.launch_inference_worker(num_api_servers)
+        self.transport_config.manager = mp.Manager()
+        self.transport_config.num_consumers = num_api_servers
+        inference_workers = []
+        for lit_api in self.litapi_connector:
+            manager, _inference_workers = self.launch_inference_worker(lit_api)
+            inference_workers.extend(_inference_workers)
```
bhimrazy, line 645: Really nice! 
I was also wondering how the queues and workers will be allocated for each of the `lit_apis`.

> I assume this might still be in progress — from what I understand, a fresh transport is created and replaces `self._transport` for each `lit_api`, so we might need to keep track of the transport (since it’s later used during the lifespan), along with any other new instances being initialized in `launch_inference_worker`.
> Please feel free to correct me if I’ve misunderstood anything!

Looking forward to the upcoming changes and learning more from them!

bhimrazy (approved):

src/litserve/server.py
aniketmaurya: great catch! yeah, we should reduce the redundancy and remove the arguments that we can fetch from LitAPI (stream, batch_size, etc.). It would be cleaner, easy to maintain and promotes high cohesion which is one of the main point of this migration.

Borda:
> TODOs:
> 
> * Detect duplicate endpoint paths
> * Don't allow streaming with non-stream

Is it for this PR or next one?

aniketmaurya:
there were in TODO in this PR itself @Borda! I just updated to resolve them.

ywh-my:
Thanks for your hard work! I can not find the branch "[endpoint-2](https://github.com/Lightning-AI/LitServe/tree/endpoint-2)" . Are you still developing it ？

aniketmaurya:
hi @ywh-my, it has been merged into the main branch. You can install the latest version of LitServe to use this feature. Here is the docs for it - https://lightning.ai/docs/litserve/features/multiple-endpoints

ywh-my:
Thank you !!!!

PR #532, opened 2025-06-09
Shutdown server when workers crash

andyland:
## What does this PR do?

Right now, if a worker crashes, the Litserve server stays up and reports it is healthy even though it is very unhealthy and can't serve requests.

Demo of the problem: https://www.loom.com/share/30443d7da3b84741b67006f6f7fbcb35

There are multiple ways this could be handled, either the service could shut down, or the workers could be restarted. Ideally, this is configured by the user.  This PR implements the first method, which I believe should be the default.  It adds a monitoring thread that polls each worker process after it is launched to see if it is still alive.   If it detects a worker has crashed, it will exit the entire server gracefully.

Demo of the solution: https://www.loom.com/share/1724eba9234a444ea9fb0adfad922c19

I will follow-up with the second method of restarting failed workers in a subsequent PR.

### DIFF
--- a/src/litserve/server.py
+++ b/src/litserve/server.py
@@ -564,6 +564,7 @@ class LitServer:
         self.litapi_request_queues = {}
         self._shutdown_event: Optional[mp.Event] = None
         self.uvicorn_graceful_timeout = 30
+        self.restart_workers = False
 
         accelerator = self._connector.accelerator
         devices = self._connector.devices
@@ -835,7 +836,6 @@ class LitServer:
         self,
         manager: mp.Manager,
         uvicorn_workers: List[Union[mp.Process, threading.Thread]],
-        inference_workers: List[mp.Process],
         shutdown_reason: str = "normal",
     ):
         """Encapsulates the graceful shutdown logic."""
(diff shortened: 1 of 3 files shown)


src/litserve/server.py
```diff
@@ -1064,3 +1066,25 @@ def shutdown_api_key_auth(self, shutdown_api_key: str = Depends(oauth2_scheme)):
...
+        uvicorn_workers: List[Union[mp.Process, threading.Thread]],
+    ):
+        def monitor():
+            while not self._shutdown_event.is_set():
+                for proc in self.inference_workers:
+                    if proc.is_alive():
+                        continue
+
+                    if not self.restart_workers:
+                        logging.error(f"⚠️ Worker {proc.name} died; shutting down")
```
aniketmaurya, line 1082:
```suggestion
                        logger.error(f"⚠️ Worker {proc.name} died; shutting down")
```

src/litserve/server.py
```diff
@@ -1064,3 +1066,25 @@ def shutdown_api_key_auth(self, shutdown_api_key: str = Depends(oauth2_scheme)):
...
+                        self._perform_graceful_shutdown(manager, uvicorn_workers, f"⚠️ Worker {proc.name} died)")
+                        return
+
+                    # TODO: Handle restarting failed workers
+                time.sleep(1)
```
aniketmaurya, line 1087:
```suggestion
                time.sleep(5)
```

aniketmaurya (approved):

PR #534, opened 2025-06-09
Enable MCP server

aniketmaurya:
## What does this PR do?

Users can create MCP server using LitServe with this PR. Any AI model served using LitServe can be used as a MCP server and connect to MCP clients such as Claude desktop or IDEs like Cursor. 

**Example:** A sentiment analysis model server with ready-to-use MCP endpoint. 

```python
from pydantic import BaseModel
from litserve.mcp import MCP
import litserve as ls

class TextClassificationRequest(BaseModel):
    input: str

class TextClassificationAPI(ls.LitAPI):
    def setup(self, device):
        self.model = pipeline("sentiment-analysis", model="stevhliu/my_awesome_model", device=device)

    def decode_request(self, request: TextClassificationRequest):
        return request.input

    def predict(self, x):
        return self.model(x)

        return output[0]

if __name__ == "__main__":
    api = TextClassificationAPI(mcp=MCP(description="Text classification"))
    server = ls.LitServer(api)
    server.run(port=8000)
```

image

<details>

</details>

## PR review

## Did you have fun?


k223kim (approved):

Borda:
Maybe add why MCP is good or important to be added?

robTheBuildr (approved):

williamFalcon:
why not add this as one of the specs? @aniketmaurya ... do we need a special arg called "mcp_spec" ? 

A: specs=[OpenAI, MCP]) for example
B: mcp=MCPServer(), specs=[]

but mcp_spec feels a bit too overfit and leaky.

aniketmaurya:
@williamFalcon the reason that I didn't make MCP as one of the spec is because it has different side-effects on LitAPI and LitServer

- specs (like the OpenAI spec) affects the API endpoint, the request-response format, and the LitAPI methods.
- MCP, on the other hand, works differently. It doesn't follow the OpenAI spec. Instead, it only uses metadata like the name and description to generate an additional endpoint via MCPServer.


I think option B would suit the best - `mcp=MCP(), specs=[]`

Issue #549, opened 2025-06-14
Add support for custom output path in generate_client_file method


KAVYANSHTYAGI:
### Feature Request

The current `generate_client_file` method in `LiteServe` always writes the generated client code to `client.py` in the current directory. This can cause issues when:

- Running multiple instances in parallel (eg- in tests)
- Users want to name the file differently or store it elsewhere

### Proposed Solution

Introduce an optional `dest_path` argument in the method signature to allow specifying a custom file output location.

### Benefits

- Greater flexibility for users
- Better testability and modularity
- Avoids overwriting existing files

aniketmaurya:
Hi @KAVYANSHTYAGI, you can disable the client generation by setting `generate_client_file=False` in `LitServer.run` method. I am closing this issue since we are trying to minimize the client generation scope.

PR #550, opened 2025-06-14
Add support for custom client file paths in `generate_client_file` with improved logging and tests

KAVYANSHTYAGI:
### PR Description:

This PR enhances the generate_client_file static method in LiteServe by adding support for a user-defined destination path (dest_path). If not specified, the method defaults to generating client.py in the current working directory, preserving backward compatibility.

**Key Improvements:**

    Custom Output Path Support: Users can now specify a custom filename and path for the generated client file.

    Improved Logging: Log messages now reflect the actual path where the client file is (or would be) written, improving clarity in debugging and automation workflows.

    Documentation Update: The inline docstring and YAML-based config description have been updated to reflect the new behavior.

    New Unit Test: Adds test_generate_client_file_custom_path to validate that the generated file correctly reflects the server URL when a custom path is used.

**Test Coverage:**

    Verified that both default (client.py) and custom output paths generate a valid client with expected contents.

    Ensured backward compatibility with existing usage.

Fix:#549

### DIFF
--- a/src/litserve/server.py
+++ b/src/litserve/server.py
@@ -987,11 +987,19 @@ class LitServer:
                 self.app.add_middleware(middleware)
 
     @staticmethod
-    def generate_client_file(port: Union[str, int] = 8000):
-        dest_path = os.path.join(os.getcwd(), "client.py")
+    def generate_client_file(port: Union[str, int] = 8000, dest_path: Optional[str] = None):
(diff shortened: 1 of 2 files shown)


aniketmaurya:
Hi @KAVYANSHTYAGI, thanks for your effort but we will keep client generation as it is since it's just a guide for users that they can edit themselves.

PR #567, opened 2025-07-02
support `mcp` package  less than v1.10.0

rongfengliang:
## What does this PR do?

Fixes # (issue).

From https://github.com/modelcontextprotocol/python-sdk/pull/993, this PR removes _convert_to_content from utilities.func_metadata to support older versions of the MCP package (less than v1.10.0). Some checks will be included.


Borda (approved):

aniketmaurya:
thank you @rongfengliang 🙌 !

PR #568, opened 2025-07-04
fix: OpenAIEmbeddingSpec setup check for multi endpoint

rongfengliang:
## What does this PR do?

Fixes # (issue).

The OpenAIEmbeddingSpec currently only checks if server.lit_api is set to one. However, LitServe now supports multiple Endpoint , so this PR adds the necessary checks for multiple Endpoint.

### DIFF
--- a/src/litserve/specs/openai_embedding.py
+++ b/src/litserve/specs/openai_embedding.py
@@ -22,6 +22,7 @@ from typing import TYPE_CHECKING, Any, List, Literal, Optional, Union
 from fastapi import HTTPException, Request, Response, status
 from fastapi import status as status_code
 from pydantic import BaseModel
+
 from litserve.specs.base import LitSpec
 from litserve.utils import LitAPIStatus
 
@@ -130,20 +131,22 @@ class OpenAIEmbeddingSpec(LitSpec):
 
     def setup(self, server: "LitServer"):
         from litserve import LitAPI
+
         super().setup(server)
 
         lit_api = server.lit_api
 
         if isinstance(lit_api, LitAPI):
             self._check_lit_api(lit_api)
-        if isinstance(lit_api,list):
+        if isinstance(lit_api, list):
             for api in lit_api:
                 self._check_lit_api(api)
 
         print("OpenAI Embedding Spec is ready.")
 
-    def _check_lit_api(self,api: "LitAPI"):
+    def _check_lit_api(self, api: "LitAPI"):
         from litserve import LitAPI
+
         if inspect.isgeneratorfunction(api.predict):
             raise ValueError(
                 "You are using yield in your predict method, which is used for streaming.",
@@ -162,6 +165,7 @@ class OpenAIEmbeddingSpec(LitSpec):
                 "Please consider replacing yield with return in encode_response.\n",
                 EMBEDDING_API_EXAMPLE,
             )
+
     def decode_request(self, request: EmbeddingRequest, context_kwargs: Optional[dict] = None) -> List[str]:
         return request.input
 
--- a/src/litserve/specs/openai_embedding.py
+++ b/src/litserve/specs/openai_embedding.py
@@ -144,7 +144,7 @@ class OpenAIEmbeddingSpec(LitSpec):
 
         print("OpenAI Embedding Spec is ready.")
 
-    def _check_lit_api(self, api: "LitAPI"):
+    def _check_lit_api(self, api):
         from litserve import LitAPI
 
         if inspect.isgeneratorfunction(api.predict):
--- a/src/litserve/specs/openai_embedding.py
+++ b/src/litserve/specs/openai_embedding.py
@@ -22,7 +22,6 @@ from typing import TYPE_CHECKING, Any, List, Literal, Optional, Union
 from fastapi import HTTPException, Request, Response, status
 from fastapi import status as status_code
 from pydantic import BaseModel
-
 from litserve.specs.base import LitSpec
 from litserve.utils import LitAPIStatus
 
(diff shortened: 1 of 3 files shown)


aniketmaurya:
hi @rongfengliang, thank you for creating the PR! Moving with version `0.3.0`, we are going to change how we pass specs to the server so one LitAPI will connect to a single spec. 

New:
```python
api = EmbeddingLitAPI(spec=ls.OpenAIEmebeddingSpec())
server = ls.Server(api)
```


Old:
```python
api = EmbeddingLitAPI()
server = ls.Server(api, spec=ls.OpenAIEmebeddingSpec())
```

rongfengliang:
@aniketmaurya  that's true, but currently  OpenAIEmbeddingSpec in setup check use server.lit_api , lit_api maybe a  list, this pr will  check if  server.lit_api  is one or multi  
```code
    def setup(self, server: "LitServer"):
        from litserve import LitAPI

        super().setup(server)

        lit_api = server.lit_api  # if lit_api is a list  below lit_api.predict will throw one exception 
        if inspect.isgeneratorfunction(lit_api.predict):
```
also Old design maybe not  very good. LitAPI use LitSpec maybe LitSpec also add LitAPI instance so ,LitSpec will not use server instance,  this will be better for checking some functions (like OpenAIEmbeddingSpec check async or sync code  )

aniketmaurya:
@rongfengliang cool let's do this for now. Can you please add a unit test [here](https://github.com/Lightning-AI/LitServe/blob/main/tests/unit/test_openai_embedding.py) too? Maybe pass litapi both as a list and a single instance to LitServe and `assert` it works both time.

src/litserve/specs/openai_embedding.py
```diff
@@ -135,7 +135,19 @@ def setup(self, server: "LitServer"):
         super().setup(server)
 
         lit_api = server.lit_api
-        if inspect.isgeneratorfunction(lit_api.predict):
+
+        if isinstance(lit_api, LitAPI):
+            self._check_lit_api(lit_api)
+        if isinstance(lit_api, list):
```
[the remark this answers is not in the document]
rongfengliang: No, can only one, because check method raise exception , write with this code will be more clear

rongfengliang:
@aniketmaurya I had added the  Unit test

tests/unit/test_openai_embedding.py
```diff
@@ -30,6 +31,7 @@
     TestEmbedAPIWithUsage,
     TestEmbedAPIWithYieldEncodeResponse,
     TestEmbedAPIWithYieldPredict,
+    TestOpenAPI,
```
aniketmaurya, line 34:
```suggestion
```

tests/unit/test_openai_embedding.py
```diff
@@ -51,6 +51,24 @@ async def test_openai_embedding_spec_with_single_input(openai_embedding_request_
...
+async def test_openai_embedding_spec_with_multi_endpoint(openai_embedding_request_data):
+    server = ls.LitServer([
+        TestEmbedAPI(spec=OpenAIEmbeddingSpec(), api_path="/v2/embeddings"),
+        TestEmbedAPI(spec=OpenAIEmbeddingSpec()),
+    ])
+    with wrap_litserve_start(server) as server:
+        async with LifespanManager(server.app) as manager, AsyncClient(
+            transport=ASGITransport(app=manager.app), base_url="http://test"
+        ) as ac:
+            resp = await ac.post("/v1/embeddings", json=openai_embedding_request_data, timeout=10)
```
aniketmaurya, line 64:
```suggestion
            resp = await ac.post("/v2/embeddings", json=openai_embedding_request_data, timeout=10)
```

tests/unit/test_openai_embedding.py
```diff
@@ -51,6 +51,24 @@ async def test_openai_embedding_spec_with_single_input(openai_embedding_request_
             assert len(resp.json()["data"][0]["embedding"]) == 768, "Embedding length should be 768"
 
 
+@pytest.mark.asyncio
+async def test_openai_embedding_spec_with_multi_endpoint(openai_embedding_request_data):
+    server = ls.LitServer([
+        TestEmbedAPI(spec=OpenAIEmbeddingSpec(), api_path="/v2/embeddings"),
+        TestEmbedAPI(spec=OpenAIEmbeddingSpec()),
+    ])
```
aniketmaurya, line 59:
```suggestion
    server = ls.LitServer([
        TestEmbedAPI(spec=OpenAIEmbeddingSpec()),
    ])
```

tests/unit/test_openai_embedding.py
```diff
@@ -51,6 +51,23 @@ async def test_openai_embedding_spec_with_single_input(openai_embedding_request_
...
+@pytest.mark.asyncio
+async def test_openai_embedding_spec_with_multi_endpoint(openai_embedding_request_data):
+    server = ls.LitServer([
+        TestEmbedAPI(spec=OpenAIEmbeddingSpec()),
+    ])
+    with wrap_litserve_start(server) as server:
+        async with LifespanManager(server.app) as manager, AsyncClient(
+            transport=ASGITransport(app=manager.app), base_url="http://test"
+        ) as ac:
+            resp = await ac.post("/v2/embeddings", json=openai_embedding_request_data, timeout=10)
```
bhimrazy, line 63:
```suggestion
            resp = await ac.post("/v1/embeddings", json=openai_embedding_request_data, timeout=10)
```

bhimrazy:
I was thinking a bit more about this, and something came to mind — ideally, it might be cleaner if the spec only needs to check the specific LitAPI instance it's attached to, rather than handling a list. Maybe this could be abstracted better via the connector.

cc: @aniketmaurya

PR #573, opened 2025-07-08
refactor:  validation logic to pre_setup method in Embed Spec, where there is access to correct api instance

bhimrazy:
## What does this PR do ?

Move validation checks for the API instance to the pre_setup method to ensure access to the correct API instance before setup. This change improves the structure and clarity of the code as similar to OpenAI chat spec.

Follow up for #568
Prep for #570 

Motivation
>  ideally, it might be cleaner if the spec only needs to check the specific LitAPI instance it's attached to, rather than handling a list. in [#568 (comment)](https://github.com/Lightning-AI/LitServe/pull/568#issuecomment-3046001009)

### DIFF
--- a/tests/unit/test_batch.py
+++ b/tests/unit/test_batch.py
@@ -170,12 +170,12 @@ def test_batch_predict_string_warning():
 
     mock_input = torch.tensor([[1.0], [2.0]])
 
+    # Simulate the behavior in run_batched_loop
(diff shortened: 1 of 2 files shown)


Borda (approved):

aniketmaurya (approved):
Great work @bhimrazy!

PR #641, opened 2025-12-03
chore: drop support for Python 3.9

bhimrazy:
## What does this PR do?

Drops support for Python 3.9 and updates the minimum required Python version to 3.10.
> Python 3.9 reached status end-of-life as of oct. 25 so lets remove support for it. ref: https://devguide.python.org/versions/

---
**Follow-ups**

- Re-enable UP045 (non-pep604-annotation-optional) and UP007 (non-pep604-annotation-union) lint rules to upgrade to Python 3.10+ syntax (X | None, X | Y)
- Add Python 3.12 and newer to CI matrix (currently commented as TODO)

## PR review

## Did you have fun?


andyland:
> I would suggest to drop the tests but allow the package installation for 3.9.
> 
> I know some companies have constraints on upgrading Python versions etc. so essentially we will be stooping them from using the latest version of LitServe. And this way we can say that, you can try it but not tested so better upgrade.
> 
> > For example, FastAPI is still to drop Python 3.8 support even after 1 year of EOL.
> 
> I'd invite maintainers to chime in and put their opinion.

I don't think we want to suggest compatibility with something we don't test and explicitly don't support

PR #682, opened 2026-05-10
fix: wait for worker setup complete in wrap_litserve_start

### LINKED ISSUE #681: wrap_litserve_start() hangs on requests when worker setup() is slow
## 🐛 Bug

`wrap_litserve_start()` is a pytest utility that starts the LitServe server in a context manager for testing. It launches inference worker processes but yields immediately without waiting for them to finish `setup()` (model loading, etc.).

This causes `TestClient.post()` to hang indefinitely because the worker hasn't started listening on the request queue yet. The request gets enqueued but never processed.

### To Reproduce

```python
import litserve as ls
from fastapi.testclient import TestClient


class SlowSetupAPI(ls.LitAPI):
    def setup(self, device):
        import time
        time.sleep(5)  # simulate slow model loading
        self.model = lambda x: x

        return request["input"]

    def predict(self, x):
        return self.model(x)

        return {"output": output}


def test_predict():
    api = SlowSetupAPI()
    server = ls.LitServer(api, accelerator="cpu")

    with ls.utils.wrap_litserve_start(server):
        client = TestClient(server.app)
        # This hangs forever because the worker is still in setup()
        response = client.post("/predict", json={"input": "hello"})

    assert response.status_code == 200
```

#### Code sample

The root cause is in `litserve/utils.py`. `wrap_litserve_start` calls `launch_inference_worker()` but never waits for `workers_setup_status` to reach `READY`, unlike `LitServer.run()` which has this loop:

```python
# From server.py run() — this wait is missing from wrap_litserve_start
while not all(v == WorkerSetupStatus.READY for v in self.workers_setup_status.values()):
    if any(v == WorkerSetupStatus.ERROR for v in self.workers_setup_status.values()):
        raise RuntimeError("One or more workers failed to start. Shutting down LitServe")
    time.sleep(0.05)
```

### Expected behavior

`wrap_litserve_start` should wait for all workers to be ready before yielding, matching the behavior of `LitServer.run()`. This would ensure `TestClient` requests are processed immediately without hanging.

### Environment

- LitServe Version: 0.2.17
- OS: Linux
- Python version: 3.12
- How you installed: `pip install litserve`

### Reproduction setup with uv

```bash
uv init --python 3.12
uv add "litserve==0.2.17" pytest pytest-timeout httpx
uv run pytest test_bug.py -v --timeout=30
```

### Additional context

**Related issues:** #263 (setup not awaited), #663 (health check returns 200 when workers not ready), #660 (stale worker status).

**Workaround:**

```python
import time

with ls.utils.wrap_litserve_start(server):
    while not all(v == "ready" for v in server.workers_setup_status.values()):
        time.sleep(0.5)
    with TestClient(server.app) as client:
        response = client.post("/predict", json={"input": "hello"})
```

**Suggested fix:** Add the worker readiness wait loop to `wrap_litserve_start` before yielding:

```python
@contextmanager
def wrap_litserve_start(server: "LitServer", worker_monitor: bool = False):
    # ... existing setup code ...

    # Wait for all workers to be ready (same as LitServer.run)
    while not all(
        v == WorkerSetupStatus.READY for v in server.workers_setup_status.values()
    ):
        if any(v == WorkerSetupStatus.ERROR for v in server.workers_setup_status.values()):
            raise RuntimeError("One or more workers failed to start")
        time.sleep(0.05)

    try:
        yield server
    finally:
        # ... existing teardown code ...
```

hesam-oxe:
## What does this PR do?

Prevents `TestClient.post()` hangs when worker `setup()` is slow — as an **opt-in** flag on `wrap_litserve_start()`.

## Why

`wrap_litserve_start()` launched inference workers but never waited for them to finish `setup()`. Tests that synchronously call `TestClient.post()` against a slow-loading model could hit a not-yet-listening worker. `LitServer.run()` already waits (via `verify_worker_status()`) — this PR brings the same capability to the pytest context manager.

## How

`wrap_litserve_start(server, worker_monitor=False, wait_for_workers=True)`:

- waits for all workers to reach `WorkerSetupStatus.READY` before yielding
- raises `RuntimeError` if any worker reports a setup error, or if the shutdown event fires mid-wait

The default (`wait_for_workers=False`) is unchanged from upstream: the context manager yields immediately and `/health` reports 503 until workers become ready — the `test_workers_health*` suite depends on that transition, which is why the earlier unconditional wait broke CI (`assert 200 == 503`).

## Tests

- `test_wrap_litserve_start_wait_for_workers[True/False]` — new: pins that the opt-in wait yields only after every worker finished `setup()` (health is 200 immediately).
- `test_workers_health[True/False]` and `test_workers_health_custom_path[True/False]` — pass again with default flags.

Verified locally: 21 passed across `tests/unit/test_simple.py` + `test_multiple_endpoints.py` (the 4 remaining collection errors are a local pytest/flaky plugin version mismatch, not present in CI).

### DIFF
--- a/src/litserve/utils.py
+++ b/src/litserve/utils.py
@@ -115,8 +115,19 @@ def wrap_litserve_start(server: "LitServer", worker_monitor: bool = False):
     else:
         server.mcp_server = None
 
-    try:
-        yield server
+    else:
+        server.mcp_server = None
+
+        # Wait for all workers to be ready
+        while not all(
+            v == WorkerSetupStatus.READY for v in server.workers_setup_status.values()
+        ):
+            if any(v == WorkerSetupStatus.ERROR for v in server.workers_setup_status.values()):
+                raise RuntimeError("One or more workers failed to start")
+            time.sleep(0.05)
+
+        try:
+            yield server
     finally:
         server._shutdown_event.set()
         # First close the transport to signal to the response_queue_to_buffer task that it should stop


hesam-oxe:
Fixed and rebased onto current `main` — ready for re-review! 🙏

What changed since the last CI run:
- **Repaired a SyntaxError** in the original hunk: the worker-readiness wait loop was accidentally nested inside a duplicated `else:` block, so `utils.py` could not even be imported (that was the CI failure). The wait now runs after MCP setup and before `yield`, and the file compiles cleanly.
- **Merged latest `main`** (30 files) with no conflicts, so the stale pre-commit/CI results are superseded by a fresh run.

hesam-oxe:
CI green (3/3), rebased on current main, SyntaxError fixed. Ready for review when you have bandwidth.

hesam-oxe:
Hi @andyland @dmitsf — this PR is CI-green (3/3) and ready whenever you have a moment. Happy to address anything. 🙏

PR #707, opened 2026-06-07
fix(logging): use module loggers instead of print()

racinmat:
## What does this PR do?

Fixes #696, fixes #697.

<details>
  <summary><b>Before submitting</b></summary>

- [ ] Was this discussed/agreed via a Github issue? (no need for typos and docs improvements)
  - [nobody reacted to them.
- [x] Did you read the [contributor guideline](https://github.com/Lightning-AI/pytorch-lightning/blob/main/.github/CONTRIBUTING.md), Pull Request section?
- [ ] Did you make sure to update the docs?
  - there is nothing to update
- [ ] Did you write any new necessary tests?
  - no, are they needed?
</details>

## PR review

Anyone in the community is free to review the PR once the tests have passed.
If we didn't discuss your PR in GitHub issues there's a high chance it will not be merged.

## Did you have fun?

Make sure you had fun coding 🙃

PR #710, opened 2026-06-12
fix(server): print localhost Swagger URL

### LINKED ISSUE #697: invalid ip address for windows during local debugging
## 🐛 Bug

0.0.0.0 ip address doesn't work on windows. It should not be shown to user.

### To Reproduce

Run the code sample.

#### Code sample

```python
import os
import sys
import threading
import time

class _MinimalAPI(LitAPI):
    def setup(self, device: str) -> None:
        self.model = lambda x: x * 2

    def decode_request(self, request: dict) -> float:
        return float(request["input"])

    def predict(self, x: float) -> float:
        return self.model(x)

    def encode_response(self, output: float) -> dict:
        return {"output": output}

if __name__ == "__main__":
    print(f"[MAIN] PID={os.getpid()}  platform={sys.platform}")
    print("[MAIN] Starting LitServer...")
    server = LitServer(
        _MinimalAPI(),
        accelerator="cpu",
        devices=1,
        workers_per_device=1,
    )
    def _prompt():
        time.sleep(4)  # wait for uvicorn + inference workers to be ready
        print("[MAIN] Server fully running — click Stop in PyCharm to reproduce the bug.")

    threading.Thread(target=_prompt, daemon=True).start()
    server.run(port=8765, generate_client_file=False, log_level="info")
```

### Expected behavior

```
Swagger UI is available at http://localhost:8765/docs
```
to be printed.

Instead, this shows
```
Swagger UI is available at http://0.0.0.0:8765/docs
```
Which doesn't work on windows.

### Environment

- PyTorch/Jax/Tensorflow Version (e.g., 1.0): -
- OS (e.g., Linux): Windows
- How you installed PyTorch (`conda`, `pip`, source): - 
- Build command you used (if compiling from source): - 
- Python version: - 
- CUDA/cuDNN version: -
- GPU models and configuration: -
- Any other relevant information: -

sarvesh1327:
## Summary
- Print a browser-usable Swagger UI URL when LitServe binds to wildcard hosts.
- Map `0.0.0.0` and `::` to `localhost` for the user-facing docs link while keeping the actual bind host unchanged.
- Add regression coverage for the display-host helper and the startup message.

## Test Plan
- [x] `.venv/bin/python -m pytest tests/unit/test_lit_server.py::test_disable_openapi_url_print_message -q` (failed before the fix with `http://0.0.0.0:8000/docs`, passed after)
- [x] `.venv/bin/python -m pytest tests/unit/test_lit_server.py::test_display_host_for_url_uses_browser_address tests/unit/test_lit_server.py::test_disable_openapi_url_print_message -q`
- [x] `uv run --with ruff ruff check src/litserve/server.py tests/unit/test_lit_server.py`

Fixes #697

### DIFF
--- a/src/litserve/server.py
+++ b/src/litserve/server.py
@@ -77,6 +77,12 @@ LIT_SERVER_API_KEY = os.environ.get("LIT_SERVER_API_KEY")
 SHUTDOWN_API_KEY = os.environ.get("LIT_SHUTDOWN_API_KEY")
 oauth2_scheme = OAuth2PasswordBearer(tokenUrl="token")
 
+
+def _display_host_for_url(host: str) -> str:
+    """Return a browser-usable host for user-facing URLs."""
+    return "localhost" if host in {"0.0.0.0", "::"} else host
+
+
 # FastAPI writes form files to disk over 1MB by default, which prevents serialization by multiprocessing
 MultiPartParser.max_file_size = sys.maxsize
 # renamed in PR: https://github.com/encode/starlette/pull/2780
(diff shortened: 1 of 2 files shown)


bhimrazy:
Closed by #707

PR #712, opened 2026-06-13
Add per-LitAPI health check endpoints

### LINKED ISSUE #605: Access Health check endpoint for each LitAPI independently
I want to be able to check health of each LitAPI separately
```md
1. localhost:8000/LITAPI_NAME_1/health
2. localhost:8000/LITAPI_NAME_1/health
```
_Originally posted by @aniketmaurya in https://github.com/Lightning-AI/LitServe/pull/604#discussion_r2334329463_

discobot:
## What does this PR do?

Fixes #605.

Each LitAPI now exposes its own health check at `{api_path}{healthcheck_path}` — e.g. `/api1/health`, `/api2/health` (and `/api1/my_server/health` with a custom `healthcheck_path`). The per-API check reports worker readiness for that API's workers only, plus its user-defined `health()` hook, returning the same `200 "ok"` / `503 "not ready"` contract as the global endpoint. Auth matches the API's predict route (`setup_auth(lit_api)`).

The per-API logic lives in a single helper, `LitServer._check_lit_api_health`, and the global healthcheck endpoint now aggregates it across all APIs, so the two can't drift. Worker keys are matched by exact `{endpoint}_{worker_id}` parsing rather than a prefix check so an API named `predict` doesn't pick up `predict_v2`'s workers. Global endpoint behavior is unchanged, including 503 on empty `workers_setup_status` and calling the `health()` hook even before workers are ready.

Two existing tests stubbed `workers_setup_status` with a synthetic `"worker-0"` key, a format `launch_inference_worker` never produces; they now use the real `"predict_0"` format. New tests cover a two-API server (healthy API 200, unhealthy API 503, global 503) and the custom `healthcheck_path` + custom `api_path` combination, parametrized over zmq like the neighboring health tests.

<details>

</details>

As a user running multiple LitAPIs behind one server, I can now ask "is *this specific* API healthy?" by hitting `/api1/health` instead of only the aggregate `/health`, which masks which API is degraded. This makes per-API liveness/readiness probes (Kubernetes, load balancers) and targeted alerting possible without custom workaround endpoints.

## PR review

## Did you have fun?

Yes 🙃

---

🤖 Implemented with the help of [Claude Code](https://claude.com/claude-code).

### DIFF
--- a/tests/unit/test_simple.py
+++ b/tests/unit/test_simple.py
@@ -177,6 +177,68 @@ def test_workers_health_with_async_health_method(use_zmq):
         assert response.text == "ok"
 
 
+class UnhealthyLitAPI(SimpleLitAPI):
+    def health(self) -> bool:
+        return False
+
+
+@pytest.mark.parametrize("use_zmq", [True, False])
+def test_per_api_health(use_zmq):
+    server = LitServer(
+        [SimpleLitAPI(api_path="/healthy"), UnhealthyLitAPI(api_path="/unhealthy")],
+        accelerator="cpu",
+        devices=1,
+        timeout=5,
+        fast_queue=use_zmq,
+    )
+
+    with wrap_litserve_start(server) as server, TestClient(server.app) as client:
+        # wait for workers to be ready
+        for _ in range(10):
+            response = client.get("/healthy/health")
+            if response.status_code == 200:
+                break
+            time.sleep(0.5)
+        assert response.status_code == 200
+        assert response.text == "ok"
+
+        response = client.get("/unhealthy/health")
+        assert response.status_code == 503
+        assert response.text == "not ready"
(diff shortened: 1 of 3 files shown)


tests/unit/test_simple.py
```diff
@@ -177,6 +177,68 @@ def test_workers_health_with_async_health_method(use_zmq):
         assert response.text == "ok"
 
 
+class UnhealthyLitAPI(SimpleLitAPI):
```
deependujha, line 180: rather than using a consistent UnhealthyLitAPI, LitAPI with deterministic health (True -> False -> True) etc and then ensuring `/health` is fine would imo be better test than with one always return False.

deependujha (approved):

Issue #696, opened 2026-06-06
Print used in multiple places instead of logger


racinmat:
## 🐛 Bug

Print is used on multiple places, making it impossible to use custom logging handlers to process the data.

### To Reproduce

Run arbitrary code.
See parts printed.

### Environment
If you published a Studio with your bug report, we can automatically get this information. Otherwise, please describe:   

- PyTorch/Jax/Tensorflow Version (e.g., 1.0):
- OS (e.g., Linux):
- How you installed PyTorch (`conda`, `pip`, source):
- Build command you used (if compiling from source):
- Python version:
- CUDA/cuDNN version:
- GPU models and configuration:
- Any other relevant information:

PR #713, opened 2026-06-15
fix(callbacks): log metric callbacks

sarvesh1327:
## Summary
- Replaces the default metric callback `print()` calls with module logger calls.
- Keeps prediction timing and active-request messages visible, while allowing logging handlers/configuration to capture and route them.
- Updates callback tests to assert log records for the direct callback paths and preserves the integration stdout assertion for the worker-process path.

## Test Plan
- [x] `uv run python -m pytest tests/unit/test_callbacks.py -q`
- [x] `uv run ruff check src/litserve/callbacks/defaults/metric_callback.py tests/unit/test_callbacks.py`
- [x] `uv run python -m py_compile src/litserve/callbacks/defaults/metric_callback.py tests/unit/test_callbacks.py`
- [x] `git diff --check`

Refs #696

### DIFF
--- a/src/litserve/callbacks/defaults/metric_callback.py
+++ b/src/litserve/callbacks/defaults/metric_callback.py
@@ -1,3 +1,4 @@
+import logging
 import time
 import typing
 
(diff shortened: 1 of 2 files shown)


bhimrazy:
Thanks for the contribution, @sarvesh1327!  We merged the changes in #707, so I'm closing this PR as it's now a duplicate.

We really appreciate your contribution and look forward to seeing more from you. Feel free to pick up any other open issues—we'd be happy to review your future PRs!

cc: @andyland

PR #725, opened 2026-08-04
Feature/prometheus metrics

### LINKED ISSUE #724: Prometheus & OpenTelemetry Metrics for LitServe
## Problem

LitServe currently provides a foundation for logging via `loggers.py`, but production-ready Prometheus metrics are left to the user. Since LitServe is a multi-process system (multiple API servers + multiple inference workers), implementing metrics correctly is non-trivial.

### Current State

LitServe handles logging through the `Logger` base class and `_LoggerConnector`.

**Architecture**

* A dedicated logger process is started when `loggers=[MyLogger()]` is passed to `LitServer`.
* A `multiprocessing.Queue` is created.
* Inside `LitAPI.predict()`, calling `self.log("metric", value)` pushes `(metric, value)` into the queue.
* The logger process consumes the queue and calls `logger.process(metric, value)`.
* The `Logger.mount(path, app)` hook allows mounting an ASGI app (e.g. a `/metrics` endpoint) onto the main FastAPI application.

### What's Missing

* **No auto-instrumentation:** HTTP request count, request latency, error rate, queue depth, etc. must all be implemented manually.
* **Multiprocessing issue:** If `mount("/metrics", prometheus_app)` is used, the endpoint is served by the API server processes, while metrics updated in `logger.process()` live only in the logger process. As a result, the API servers cannot expose those metrics.
* **No standard integration:** Every user has to write their own `prometheus_client` integration and handle multiprocessing concerns themselves.

hillhack:
## Fixes

Closes #724

# Feature: Native Prometheus Metrics for Multiprocess Architecture

## Problem

LitServe provides a solid logging foundation through `LitAPI.log()`, but production users currently need to manually integrate Prometheus for request, latency, and inference metrics.

Supporting Prometheus in LitServe is challenging because metrics are produced across multiple processes:

* **HTTP metrics** are generated by the API server (Uvicorn/FastAPI).
* **Inference metrics** (`self.log()`) are sent through an `mp.Queue` and processed by a dedicated Logger process.
* The **`/metrics`** endpoint is served by the API server, but it must expose metrics generated by both the API and Logger processes.

Prometheus supports this architecture via `PROMETHEUS_MULTIPROC_DIR`, but it requires careful lifecycle management. If the shared directory is initialized after metrics are created, or initialized more than once, metrics become fragmented and the `/metrics` endpoint reports incorrect results.

---

## Solution

This PR adds native Prometheus support as an optional dependency (`pip install litserve[prometheus]`) with a multiprocess-safe implementation.

### Architecture

* **Centralized lifecycle (`server.py`)**

  * `LitServer` owns the creation of a single shared metrics directory.
  * `PROMETHEUS_MULTIPROC_DIR` is configured before any worker or logger processes are started.
  * `PrometheusMiddleware` is automatically registered when `PrometheusLogger` is enabled.

* **Dedicated `metrics.py` module**

  * `PrometheusMiddleware` automatically records:

    * `litserve_http_requests_total`
    * `litserve_http_request_duration_seconds`
  * Internal endpoints such as `/health`, `/info`, and `/metrics` are excluded from instrumentation.
  * `PrometheusLogger` processes inference metrics emitted through `self.log()` and exposes the `/metrics` ASGI endpoint.

* **Lazy metric initialization**

  * Prometheus `Counter` and `Histogram` objects are created only after the server has configured `PROMETHEUS_MULTIPROC_DIR`, preventing initialization-order race conditions in a multiprocess environment.

* **Graceful cleanup**

  * The temporary multiprocess metrics directory is removed during server shutdown to prevent stale metrics and leftover files across restarts.

---

## Testing

Added integration tests in `tests/unit/test_prometheus.py` to verify:

* Prometheus initialization and `PROMETHEUS_MULTIPROC_DIR` setup.
* HTTP request metrics are correctly recorded and exposed.
* End-to-end inference metrics flow:

  * `predict() → self.log() → Logger → /metrics`
* Cleanup of the temporary metrics directory during graceful shutdown.

### DIFF
--- a/src/litserve/__init__.py
+++ b/src/litserve/__init__.py
@@ -37,7 +37,8 @@ __all__ = [
 ]
 
 try:
-    from litserve.metrics import PrometheusLogger
+    from litserve.metrics import PrometheusLogger  # noqa: F401
+
     __all__.append("PrometheusLogger")
 except ImportError:
     pass
--- a/src/litserve/__init__.py
+++ b/src/litserve/__init__.py
@@ -35,3 +35,9 @@ __all__ = [
     "set_trace_if_debug",
     "test_examples",
 ]
+
+try:
+    from litserve.metrics import PrometheusLogger
+    __all__.append("PrometheusLogger")
+except ImportError:
+    pass
--- a/src/litserve/__init__.py
+++ b/src/litserve/__init__.py
@@ -16,6 +16,7 @@ from litserve.__about__ import *  # noqa: F403
 from litserve.api import LitAPI
 from litserve.callbacks import Callback
 from litserve.loggers import Logger
+from litserve.metrics import PrometheusLogger
 from litserve.server import LitServer, Request, Response
 from litserve.specs import OpenAIEmbeddingSpec, OpenAISpec
 from litserve.utils import configure_logging, set_trace, set_trace_if_debug
@@ -29,16 +30,10 @@ __all__ = [
     "Logger",
     "OpenAISpec",
     "OpenAIEmbeddingSpec",
+    "PrometheusLogger",
     "Request",
     "Response",
     "set_trace",
     "set_trace_if_debug",
     "test_examples",
 ]
-
-try:
-    from litserve.metrics import PrometheusLogger  # noqa: F401
-
-    __all__.append("PrometheusLogger")
-except ImportError:
-    pass
--- a/src/litserve/metrics.py
+++ b/src/litserve/metrics.py
@@ -17,16 +17,11 @@ import time
 from starlette.middleware.base import BaseHTTPMiddleware
 
 from litserve.loggers import Logger
+from litserve.utils import is_package_installed
 
 logger = logging.getLogger(__name__)
 
-try:
-    import prometheus_client  # noqa: F401
-    from prometheus_client import CollectorRegistry, Counter, Histogram, make_asgi_app, multiprocess
-
-    _PROMETHEUS_AVAILABLE = True
-except ImportError:  # pragma: no cover
-    _PROMETHEUS_AVAILABLE = False
+_PROMETHEUS_AVAILABLE = is_package_installed("prometheus_client")
 
 
 def _check_prometheus_available():
@@ -46,6 +41,8 @@ class PrometheusMiddleware(BaseHTTPMiddleware):
 
         # Use singleton pattern to avoid hitting private registry internals or re-creating metrics
         if not hasattr(PrometheusMiddleware, "_metrics_initialized"):
+            from prometheus_client import Counter, Histogram
+
             PrometheusMiddleware.http_requests_total = Counter(
                 "litserve_http_requests_total", "Total requests", ["method", "endpoint", "status"]
             )
@@ -78,6 +75,8 @@ class PrometheusLogger(Logger):
 
         async def lazy_asgi_app(scope, receive, send):
             if not hasattr(self, "_asgi_app"):
+                from prometheus_client import CollectorRegistry, make_asgi_app, multiprocess
+
                 registry = CollectorRegistry()
                 multiprocess.MultiProcessCollector(registry)
                 self._asgi_app = make_asgi_app(registry=registry)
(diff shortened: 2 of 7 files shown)


src/litserve/__init__.py
```diff
@@ -35,3 +35,10 @@
     "set_trace_if_debug",
     "test_examples",
 ]
+
+try:
+    from litserve.metrics import PrometheusLogger  # noqa: F401
+
+    __all__.append("PrometheusLogger")
+except ImportError:
+    pass
```
deependujha: imo we should either import it as it is (without try-except) or don't import at all. This seems off.

src/litserve/metrics.py
```diff
@@ -0,0 +1,99 @@
...
+
+logger = logging.getLogger(__name__)
+
+try:
+    import prometheus_client  # noqa: F401
+    from prometheus_client import CollectorRegistry, Counter, Histogram, make_asgi_app, multiprocess
+
+    _PROMETHEUS_AVAILABLE = True
+except ImportError:  # pragma: no cover
+    _PROMETHEUS_AVAILABLE = False
```
deependujha: rather than using try-except for importing at module level, let's do it within functions/methods that require it. It is anyway cached so i don't think there's any penalty to it.

src/litserve/server.py
```diff
@@ -753,6 +753,27 @@ def __init__(
...
+                    from litserve.metrics import PrometheusMiddleware
+
+                    self._metrics_dir = tempfile.mkdtemp(prefix="litserve_prom_")
+                    os.environ["PROMETHEUS_MULTIPROC_DIR"] = self._metrics_dir
+                    import prometheus_client.values
+
+                    prometheus_client.values.ValueClass = prometheus_client.values.MultiProcessValue()
+                    middlewares.append(PrometheusMiddleware)
+        except ImportError:
+            pass
```
deependujha: it'd be better as a separate helper private function than having prometheus specific code in the server

src/litserve/server.py
```diff
@@ -1234,6 +1256,25 @@ def _perform_graceful_shutdown(
...
+                    if lp.is_alive():
+                        logger.warning(f"Logger process (PID: {lp.pid}) did not terminate gracefully. Killing.")
+                        lp.kill()
+                except Exception as e:
+                    logger.error(f"Error during termination of logger process: {e}")
+
+        if getattr(self, "_metrics_dir", None) and os.path.exists(self._metrics_dir):
+            import shutil
+
+            shutil.rmtree(self._metrics_dir, ignore_errors=True)
```
deependujha: imo, `Logger` class should have a **`close`** method, and `server.py` should just call `logger.close()` and logger implementation should deal with it.

src/litserve/utils.py
```diff
@@ -120,10 +120,27 @@ def wrap_litserve_start(server: "LitServer", worker_monitor: bool = False):
     finally:
         server._shutdown_event.set()
         # First close the transport to signal to the response_queue_to_buffer task that it should stop
-        server._transport.close()
+        if hasattr(server, "_transport"):
+            server._transport.close()
```
deependujha, line 124: why this? For server, we already know that it has `_transport` attribute, then why we need another explicit check?

src/litserve/utils.py
```diff
@@ -120,10 +120,27 @@ def wrap_litserve_start(server: "LitServer", worker_monitor: bool = False):
...
+                lp.join(timeout=1)
+                if lp.is_alive():
+                    lp.kill()
+
+        if getattr(server, "_metrics_dir", None) and os.path.exists(server._metrics_dir):
+            import contextlib
+            import shutil
+
+            with contextlib.suppress(Exception):
+                shutil.rmtree(server._metrics_dir)
```
deependujha: at best, we should just call `server._logger_connector.close()` and the implementation should be concerned with LoggerConnector only

src/litserve/server.py
hillhack: That makes total sense. Having `shutil.rmtree(self._metrics_dir)` inside `server.py` leaks logger-specific cleanup details into the server shutdown logic. i  will add a `close()` lifecycle method to the `Logger` base class.

src/litserve/utils.py
hillhack: `_transport` isn't initialized inside LitServer.__init__—it gets created dynamically later during server.run() / _init_manager().

wrap_litserve_start is a test utility helper. The hasattr(server, "_transport") check in the finally: teardown block is defensive: if a test setup fails before _init_manager() finishes instantiating _transport, accessing server._transport.close() directly would throw a secondary AttributeError, which would mask the actual test exception that caused the setup failure.

PR #733, opened 2026-08-23
fix: preserve raw request bodies for custom content types

### LINKED ISSUE #732: Pass binary payloads to `decode_request` unchanged
## 🚀 Feature

Pass binary payloads to `decode_request` unchanged.

### Motivation

We use LitServe for BYOC SageMaker Inference endpoints that take protobuf requests of the following format:

```http
POST /invocations
Content-Type: application/x-recordio-protobuf

<serialized protobuf bytes>
```

LitServe currently parses all non-form/multipart request requests as JSON. To support `application/x-recordio-protobuf`, we override private methods:

```python

class _RawRequestHandler(RegularRequestHandler):
    """Read the raw protobuf body instead of parsing JSON. The returned dict (bytes + str)
    is picklable, so it crosses the worker transport cleanly."""

    async def _prepare_request(
        self, request: Request, request_type: Any
    ) -> dict[str, Any]:
        return {
            "body": await request.body(),
            "content_type": request.headers.get("Content-Type", ""),
        }


class RankerLitServer(ls.LitServer):
    def _register_api_endpoints(
        self, lit_api: ls.LitAPI, request_type: Any, response_type: Any
    ) -> None:
        handler = _RawRequestHandler(lit_api, self)

        async def endpoint_handler(request: Request) -> Response:
            return await handler.handle_request(request, Request)

        self.app.add_api_route(
            lit_api.api_path,
            endpoint_handler,
            methods=["POST"],
            dependencies=[Depends(self.setup_auth())],
        )
```

While it works, overriding private APIs makes future LitServe upgrades brittle.

### Pitch

Enable LitServe to pass bytes from the request body to `decode_request`. This can either be opt-in manually, or via introspecting the `Content-type` header. For raw byte bodies, LitServe
should:

1. Read the request via `await request.body()` without JSON parsing.
2. Pass the resulting `bytes` to `decode_request`.
3. Make the request content type available to the API.
4. Preserve the existing worker-transport and batching behavior.
5. Leave existing JSON and multipart handling unchanged by default.

### Alternatives

- **Private handler override:** This is our current solution, but it relies on
  internal methods that may change between LitServe releases.

### Additional context

Past issue #362

mikemikimike:
## Summary
- Closes #732
- Preserve raw request bodies for non-JSON, non-form content types.
- Add a regression test covering binary payloads.

## Implementation
```text
Requests with `application/json` (including `+json`) continue through JSON decoding, and form/multipart requests keep their existing behavior. Other request content types now use FastAPI's `Request.body()`, allowing `decode_request` to receive the original bytes unchanged.
```

## Compatibility
Existing JSON and form/multipart request handling is unchanged. Custom content types that previously failed JSON parsing now receive raw bytes.

## Tests
- `python -m compileall -q src/litserve/server.py tests/unit/test_request_handlers.py` — passed
- `git diff --check` — passed
- `pytest tests/unit/test_request_handlers.py -q` — not completed: the uv-managed environment remained silent/hung during dependency setup/import in this Windows environment; the process was stopped.
- Docker checks were not run; this change does not require external services and the repository provides no Docker test configuration.

### DIFF
--- a/src/litserve/server.py
+++ b/src/litserve/server.py
@@ -32,7 +32,7 @@ from collections import deque
 from collections.abc import Callable, Iterable, Mapping, Sequence
 from contextlib import asynccontextmanager
 from queue import Queue
-from typing import TYPE_CHECKING, Literal, Optional, Union
+from typing import TYPE_CHECKING, Any, Literal, Optional, Union
 
 import uvicorn
 import uvicorn.server
@@ -294,10 +294,10 @@ class BaseRequestHandler(ABC):
         self.lit_api = lit_api
         self.server = server
 
-    async def _prepare_request(self, request, request_type) -> dict:
+    async def _prepare_request(self, request, request_type) -> Any:
         """Common request preparation logic."""
         if request_type == Request:
-            content_type = request.headers.get("Content-Type", "")
+            content_type = request.headers.get("Content-Type", "").split(";", 1)[0].strip().lower()
             if content_type == "application/x-www-form-urlencoded" or content_type.startswith("multipart/form-data"):
                 return await request.form()
             if content_type == "application/json" or content_type.endswith("+json"):
--- a/src/litserve/server.py
+++ b/src/litserve/server.py
@@ -300,7 +300,9 @@ class BaseRequestHandler(ABC):
             content_type = request.headers.get("Content-Type", "")
             if content_type == "application/x-www-form-urlencoded" or content_type.startswith("multipart/form-data"):
                 return await request.form()
-            return await request.json()
+            if content_type == "application/json" or content_type.endswith("+json"):
+                return await request.json()
+            return await request.body()
         return request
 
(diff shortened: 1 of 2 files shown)


mikemikimike:
## Maintainer feedback requested: exposing Content-Type to `decode_request`

This PR now preserves raw request bytes for custom content types and covers JSON, `+json`, charset parameters, and form/multipart compatibility.

One part of #732 remains an API design question: `decode_request` receives the raw `bytes`, but it does not receive the original `Content-Type`. For protobuf and other content-type-specific decoders, callers may need that metadata.

Could maintainers confirm the preferred follow-up/API shape?

- pass a request/context object containing both body and headers;
- add an optional context parameter to `decode_request`; or
- intentionally leave content-type handling to the endpoint/API implementation.

I have not introduced a new API in this PR to avoid guessing at a compatibility-sensitive design.

PR #741, opened 2026-09-06
fix(embedding specs): return base64 embeddings when encoding_format="base64"

### LINKED ISSUE #740: Embedding responses ignore explicit base64 encoding_format
## Bug

`POST /v1/embeddings` accepts `encoding_format="base64"` but returns a JSON float array in `data[*].embedding`. The accepted value was added in #388; the response path still ignores it.

### To reproduce

Use a CPU embedding API returning deterministic vectors, such as `[5.0, -0.5, 0.25]`, and send:

```json
{"input":"hello","model":"local-embedding","encoding_format":"base64"}
```

On main at `4c9fecc2a10c71f7cf016beb67ca0cadca2cbfc0`, `data[0].embedding` is `[5.0, -0.5, 0.25]`, a JSON array. The same happens with multiple text inputs and with both multiprocessing and ZeroMQ transports.

A complete deterministic CPU reproduction is in [this test file on the prepared branch](https://github.com/emecii/LitServe/blob/8ecac32b5f6cbb3fc0d9e9ff8a94231aeb89fb96/tests/integration/test_embedding_encoding.py). Keeping that file while using the main production source, run:

```sh
python -m pytest -q tests/integration/test_embedding_encoding.py
```

Six base64 cases fail; eight omitted/float controls pass. The tests start real inference workers through the project test helper, with synthetic data and no downloaded model or credentials. A cloud Studio was not used.

### Expected behavior

An explicit base64 request returns an encoded float32 vector string, as specified by the [embeddings API](https://developers.openai.com/api/reference/resources/embeddings/methods/create). Omitted or `float` encoding continues returning float arrays. Indices, model, and usage metadata are unchanged.

### Environment

macOS 26.5.2 arm64; Python 3.12.12; editable LitServe main at the SHA above; FastAPI 0.141.1, Pydantic 2.13.5, HTTPX 0.28.1, NumPy 2.5.2. CPU only; no CUDA or GPU. The original minimal reproduction did not need Torch; neighboring suite validation later used Torch 2.14.0.

### Additional context

A focused patch is prepared using standard-library float32 packing and base64 encoding after the existing normalization/validation. It adds no runtime dependency. The regression suite passes with that patch. This report and implementation were prepared and tested with Codex assistance; no prior maintainer agreement is asserted.

emecii:
## What does this PR do?

Fixes #740.

`encoding_format="base64"` was accepted on `/v1/embeddings` but never applied — the server always returned a JSON float array. This encodes the response when base64 is requested; `float` and the default are unchanged.

- Packs each embedding as little-endian float32, which is what OpenAI clients decode
- `Embedding.embedding` becomes `Union[list[float], str]`, matching the OpenAI schema
- Standard library only, no new dependencies
- Tests added to the existing unit and e2e embedding suites

<details>
  <summary><b>Wire format vs other OpenAI-compatible servers</b></summary>

Verified byte-identical output against each of these for the same input vector:

|  | dtype | endianness |
| --- | --- | --- |
| this PR | float32 | explicit little-endian |
| SGLang | float32 | explicit little-endian |
| Ollama | float32 | explicit little-endian |
| vLLM | float32 (default) | native, little-endian in practice |

</details>


bhimrazy (approved):
Thanks @emecii for the fixes. I made a few little changes and organized the tests into the existing format.

bhimrazy:
CI failures due to mcp being resolved here: #743

tchaton (approved):

PR #748, opened 2026-09-10
fix: reset active request count after errors

### LINKED ISSUE #747: Failed requests leave active_requests permanently incremented
## Bug

With `track_requests=True`, `RequestCountMiddleware` increments the shared counter before dispatch and decrements it only after the downstream app returns. If request handling raises, the decrement is skipped. Each 500 response can therefore leave a phantom active request in `LitServer.active_requests`, affecting callbacks and monitoring that rely on the count.

## Reproduction

On `main` at `16248473bc3127bef5d2cfe11de9d9677d9ee2c3`:

```python
import multiprocessing as mp
from fastapi import FastAPI
from fastapi.testclient import TestClient
from litserve.middlewares import RequestCountMiddleware

counter = mp.Value("i", 0, lock=True)
app = FastAPI()
app.add_middleware(RequestCountMiddleware, active_counter=counter)

@app.get("/fail")
def fail():
    raise RuntimeError("boom")

with TestClient(app, raise_server_exceptions=False) as client:
    assert client.get("/fail").status_code == 500

assert counter.value == 0  # currently 1
```

## Expected behavior

The counter is decremented when downstream request handling exits, whether it returns normally or raises, while preserving the original exception behavior.

This report was prepared with Codex assistance.

emecii:
## What does this PR do?

Fixes #747.

With `track_requests=True`, `RequestCountMiddleware` decremented the active-request counter only after a successful downstream return, so a request that raised left the counter permanently incremented. Repeated failures inflate `LitServer.active_requests` with requests that are no longer running.

- Wrap the downstream dispatch in `try`/`finally` so the counter is restored whether the request succeeds or fails, with no change to exception handling
- Add a regression test that drives a real request through `LitServer` and asserts `active_requests` returns to zero

<details>
<summary>Why the test injects the failure from a middleware</summary>

A `LitAPI` that raises can't reproduce the leak. The request handlers convert every exception into an `HTTPException`, and Starlette's `ExceptionMiddleware` sits inside `RequestCountMiddleware`, so it becomes a response before it ever crosses the counter.

The test therefore uses the public `middlewares=` parameter and raises after `call_next`, so a full `/predict` round-trip completes before the error propagates through the counter. It fails on `main` (`active_requests == 1`) and passes with the fix.

</details>

## Before submitting

- [x] Bug reported in #747 before implementation
- [x] Read the contributor guidelines
- [x] Added a regression test
- [x] No documentation update needed; public behavior is unchanged apart from correcting stale counts

AI assistance was used to investigate, implement, and test this change. No human review is claimed.

### DIFF
--- a/src/litserve/middlewares.py
+++ b/src/litserve/middlewares.py
@@ -66,5 +66,7 @@ class RequestCountMiddleware(BaseHTTPMiddleware):
             return
 
         self.active_counter.value += 1
-        await self.app(scope, receive, send)
-        self.active_counter.value -= 1
+        try:
+            await self.app(scope, receive, send)
+        finally:
+            self.active_counter.value -= 1
--- a/tests/unit/test_middlewares.py
+++ b/tests/unit/test_middlewares.py
@@ -12,10 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import copy
-import multiprocessing as mp
 
 import pytest
-from fastapi import FastAPI
 from fastapi.testclient import TestClient
 from starlette.middleware.base import BaseHTTPMiddleware
 from starlette.middleware.httpsredirect import HTTPSRedirectMiddleware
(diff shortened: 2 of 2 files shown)


bhimrazy (approved):
Thanks @emecii

deependujha (approved):

dmitsf (approved):

